PR #85, opened 2022-10-17
Respond with BindingResponse w/ embedded APIServiceBindingRequests

sttts:


pkg/kubectl/bind/plugin/bind.go
```diff
@@ -207,101 +248,105 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
...
-		// check if the CRD is owner-refed by the APIServiceBinding
-		for _, ref := range crd.OwnerReferences {
-			parts := strings.SplitN(ref.APIVersion, "/", 2)
-			if parts[0] != kubebindv1alpha1.SchemeGroupVersion.Group || ref.Kind != "APIServiceBinding" {
-				continue
-			}
-
-			existing, err := bindClient.KubeBindV1alpha1().APIServiceBindings().Get(ctx, ref.Name, metav1.GetOptions{})
+	for _, request := range apiRequests {
+		for _, resource := range request.Spec.Resources {
```
s-urbaniak, line 259: follow-up: split into separate func, this gets big

s-urbaniak (approved):

pkg/kubectl/bind/plugin/bind.go
sttts: This will be split into the two commands.

But even after that, this logic needs unit tests.

PR #93, opened 2022-10-20
contrib/example-backend: remove cluster-admin permission from service account

s-urbaniak:


contrib/example-backend/controllers/clusterbinding/clusterbinding_reconcile.go
```diff
@@ -217,3 +224,43 @@ func (r *reconciler) ensureRBACClusterRoleBinding(ctx context.Context, clusterBi
...
+
+func (r *reconciler) ensureRBACRoleBinding(ctx context.Context, clusterBinding *kubebindv1alpha1.ClusterBinding) error {
+	binding, err := r.getRoleBinding("kube-binder", clusterBinding.Namespace)
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get RoleBinding \"kube-bind\": %w", err)
+	}
+
+	expected := &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "kube-binder",
```
sttts, line 236: does this equal the service-account name?

s-urbaniak: nack, it references the kubeconfigsecretref.name with the name of the SA.

Do we want to have it the same as well? sgtm actually

contrib/example-backend/deploy/01-clusterrole.yaml
```diff
@@ -0,0 +1,55 @@
...
+- apiGroups:
+    - "kube-bind.io"
+  resources:
+    - "apiserviceexports/status"
+  verbs: ["get","patch","update"]
+- apiGroups:
+    - "kube-bind.io"
+  resources:
+    - "apiservicenamespaces"
+  verbs: ["get", "watch", "list"]
```
sttts, line 40: also write

sttts: oops, this is here twice.

s-urbaniak: fixed

contrib/example-backend/deploy/bootstrap.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/kube-bind/kube-bind/pkg/bootstrap"
+)
+
+//go:embed *.yaml
+var raw embed.FS
+
+func Bootstrap(ctx context.Context, discoveryClient discovery.DiscoveryInterface, dynamicClient dynamic.Interface, batteriesIncluded sets.String) error {
+	return bootstrap.Bootstrap(ctx, discoveryClient, dynamicClient, batteriesIncluded, raw)
+}
```
sttts, line 35: intentional here?

sttts: nvmd. It is.

s-urbaniak: yes, we want to bootstrap the cluster-role so it is known, it is static. it then will be referenced as a role-binding during reconcilation.

sttts (approved):

PR #97, opened 2022-10-21
Add kube-bind namespace UID as clusterID

sttts:
And clean up auth types.


pkg/kubectl/bind/plugin/bind.go
```diff
@@ -167,7 +166,20 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
...
+		ns = &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "kube-bind",
+			},
+		}
+		if ns, err = kubeClient.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{}); err != nil {
+			return err
+		}
+	}
+	if err := b.authenticate(provider, auth.Endpoint(ctx), sessionID, string(ns.UID), urlCh); err != nil {
```
s-urbaniak, line 182: neat idea about `ns.UID`. not sure about the guarantee of a global conflict for all UIDs but good enough for now.

s-urbaniak (approved):

pkg/kubectl/bind/plugin/bind.go
sttts: we still use the oidc auth id. So it only has to be unique for that user.

PR #98, opened 2022-10-21
kubectl-bind: fix request loop when pre-existing

sttts:
We left the loop early and did not return any binding to print.

### DIFF
--- a/pkg/kubectl/bind-apiservice/plugin/servicebindingrequest.go
+++ b/pkg/kubectl/bind-apiservice/plugin/servicebindingrequest.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
 
 	kubebindv1alpha1 "github.com/kube-bind/kube-bind/pkg/apis/kubebind/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/kubectl/bind-apiservice/plugin/servicebindingrequest.go
```diff
@@ -103,8 +107,10 @@ func (b *BindAPIServiceOptions) createAPIServiceBindings(ctx context.Context, co
 				}
 
 				if existing.Spec.KubeconfigSecretRef.Namespace == "kube-bind" && existing.Spec.KubeconfigSecretRef.Name == secretName {
+					fmt.Fprintf(b.Options.IOStreams.ErrOut, "Updating kubeconfig in secret kube-bind/%s for existing APIServiceBinding %s.\n", secretName, existing.Name) // nolint: errcheck
 					_, err = b.ensureKubeconfigSecretWithLogging(ctx, kubeconfig, secretName, kubeClient)
-					return nil, err
+					bindings = append(bindings, existing)
+					continue nextRequest
```
s-urbaniak, line 113: the label is so many lines above 😱

s-urbaniak (approved):

pkg/kubectl/bind-apiservice/plugin/servicebindingrequest.go
sttts: this function needs a refactor AND a unit test.

PR #108, opened 2022-10-22
support oidc-authorize-url

moadqassem:


contrib/example-backend/http/handler.go
```diff
@@ -75,6 +77,7 @@ func NewHandler(
 	scope kubebindv1alpha1.Scope,
 	mgr *kubernetes.Manager,
 	apiextensionsLister apiextensionslisters.CustomResourceDefinitionLister,
+	oidcAuthorizeURL string,
```
sttts, line 80: why not keep the order? Put it behind the callback URL everywhere and name both similarly.

contrib/example-backend/http/handler.go
```diff
@@ -67,6 +67,8 @@ type handler struct {
 	apiextensionsLister apiextensionslisters.CustomResourceDefinitionLister
 
 	kubeManager *kubernetes.Manager
+
+	oidcAuthorizeURL string
```
sttts, line 71: this one too

sttts (approved):

contrib/example-backend/http/handler.go
```diff
@@ -99,6 +101,10 @@ func (h *handler) AddRoutes(mux *mux.Router) {
 func (h *handler) handleServiceExport(w http.ResponseWriter, r *http.Request) {
 	logger := klog.FromContext(r.Context()).WithValues("method", r.Method, "url", r.URL.String())
 
+	if h.oidcAuthorizeURL == "" {
+		h.oidcAuthorizeURL = fmt.Sprintf("http://%s/authorize", r.Host)
```
sttts, line 105: you cannot do this. This panics due a data race.

PR #111, opened 2022-10-23
bind: considerably simplify binding creation

sttts:
Now that APIServiceExport is simpler and APIServiceBindings are named as CRDs.


s-urbaniak (approved):
Awesome

PR #113, opened 2022-10-23
contrib/example-backend: secure cookies

s-urbaniak:


contrib/example-backend/options/cookie.go
```diff
@@ -0,0 +1,74 @@
...
+	SigningKey    string
+	EncryptionKey string
+}
+
+func NewCookie() *Cookie {
+	return &Cookie{}
+}
+
+func (options *Cookie) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&options.SigningKey, "cookie-signing-key", options.SigningKey, "The key which is used to sign cookies, base64 encoded. Valid lengths are 32 or 64 bytes.")
```
sttts, line 36: as what? base64?

sttts: this should rather be a file for security, i.e. `cookie-signing-key-file`.

sttts: this should rather be a file for security, i.e. `cookie-signing-key-file`.

s-urbaniak: yes 👍 it's in the option help `The key which is used to sign cookies, base64 encoded.`

sttts (approved):

contrib/example-backend/options/cookie.go
s-urbaniak: > file for security

let's do this as a follow-up, there is prior art which uses CLI options. Having said that, I'd like to remove the cookie handling at all in a refactoring.

PR #129, opened 2022-10-29
*: enable client-go auth plugins

sttts:
Fixing https://twitter.com/JamesMunnelly/status/1585957861359726592.

### DIFF
--- a/pkg/kubectl/bind/cmd/cmd.go
+++ b/pkg/kubectl/bind/cmd/cmd.go
@@ -24,6 +24,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/exec"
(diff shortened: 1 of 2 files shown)


sttts:
@munnerz ptal.

pkg/kubectl/bind/cmd/cmd.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
```
liggitt, line 26: if you narrowly include the exec and oauth plugins instead of the "include all" package, you'll avoid pulling in the deprecated azure and gcp ones that are gone in 1.26, and keep your dep graph much smaller

sttts: Oh, perfect. Thanks for the hint. Updating.

moadqassem (approved):
LGTM

PR #131, opened 2022-10-31
Refactor the backend HTTP Server

moadqassem:
This is a bit of plumbing for the http server, in order to use it instead of using different libs(`"github.com/labstack/echo/v4"` ) for achieving similar functionalities.

### DIFF
--- a/contrib/example-backend/http/handler.go
+++ b/contrib/example-backend/http/handler.go
@@ -29,7 +29,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gorilla/mux"
 	"github.com/gorilla/securecookie"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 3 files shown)


contrib/example-backend/http/server.go
```diff
@@ -52,26 +52,31 @@ func NewServer(options *options.Serve) (*Server, error) {
...
-			server.ServeTLS(s.listener, s.options.CertFile, s.options.KeyFile) // nolint:errcheck
+			s.server.ServeTLS(s.listener, s.options.CertFile, s.options.KeyFile) // nolint:errcheck
 		}
 	}()
 
 	return nil
 }
+
+func (s *Server) Shutdown() error {
+	return s.server.Close()
```
s-urbaniak, line 81: do you mind to remove this in favor of https://github.com/kube-bind/kube-bind/pull/131/commits/31aa349c6f3cf9d4495a9c09d4158aa8cb90bfe3#diff-fee0857912127761b01d43d3a59e2f9e5c226206cc14c38b12da75f96617a798R84-R86 so it's consistent in git history?

sttts: am not convinced of turning our server into an `http.Server`. We should abstract, not inherit.

contrib/example-backend/http/handler.go
```diff
@@ -99,12 +98,12 @@ func NewHandler(
 	}, nil
 }
 
-func (h *handler) AddRoutes(mux *mux.Router) {
-	mux.HandleFunc("/export", h.handleServiceExport).Methods("GET")
-	mux.HandleFunc("/resources", h.handleResources).Methods("GET")
-	mux.HandleFunc("/bind", h.handleBind).Methods("GET")
-	mux.HandleFunc("/authorize", h.handleAuthorize).Methods("GET")
-	mux.HandleFunc("/callback", h.handleCallback).Methods("GET")
+func (h *handler) AddRoutes(server *Server) {
```
sttts, line 101: wrong direction. Handlers should not know about a server they are added to.

PR #134, opened 2022-11-07
contrib/example-backend: remove listen-ip/listen-port in favor of listen-address

s-urbaniak:
This introduces `--listen-address` which is more idiomatic and can include a port as well.
This is a preparation for new command line options for the cache implementation.

cc @sttts

### DIFF
--- a/contrib/example-backend/options/serve.go
+++ b/contrib/example-backend/options/serve.go
@@ -26,6 +26,7 @@ import (
 type Serve struct {
 	ListenIP          string
 	ListenPort        int
+	ListenAddress     string
 	CertFile, KeyFile string
 
 	// Listener is used to pre-wire a port zero listener for testing.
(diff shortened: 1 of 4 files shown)


moadqassem:
@s-urbaniak seems that import grouping needs to be adjusted

contrib/example-backend/options/serve.go
```diff
@@ -34,25 +35,25 @@ type Serve struct {
...
 	}
 }
 
 func (options *Serve) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&options.ListenIP, "listen-ip", options.ListenIP, "The host IP where the backend is running")
-	fs.IntVar(&options.ListenPort, "listen-port", options.ListenPort, "The host port where the backend is running")
-	fs.StringVar(&options.CertFile, "tls-cert-file", options.CertFile, "The TLS certificate file the webserver will use")
-	fs.StringVar(&options.KeyFile, "tls-key-file", options.KeyFile, "The TLS private key file the webserver will use")
+	fs.StringVar(&options.ListenIP, "listen-ip", options.ListenIP, "The host IP where the backend is running (DEPRECATED)")
+	fs.IntVar(&options.ListenPort, "listen-port", options.ListenPort, "The host port where the backend is running (DEPRECATED)")
```
sttts, line 44: pflag has a deprecation method.

s-urbaniak: fixed

sttts (approved):

PR #135, opened 2022-11-08
Seed rng for random oauth sessionid

rikatz:
The random number generator is not being properly initialized. This makes the sessionID on oauth workflow not be properly randomized, and make it easier to intercept/impersonate it.


sttts:
Looks great. Thank you for the contribution!

PR #137, opened 2022-11-14
Use BindOptions-wide resource printer

ardaguclu:
ResourcePrinter supports adding yaml separator automatically. Only requirement is using the same ResourcePrinter object for all resources.

This PR removes the hacky solution and makes changes to better align with standard kubectl commands.

### DIFF
--- a/pkg/kubectl/bind/plugin/bind.go
+++ b/pkg/kubectl/bind/plugin/bind.go
@@ -39,8 +39,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/component-base/logs"
 	logsv1 "k8s.io/component-base/logs/api/v1"
@@ -55,8 +55,9 @@ type BindOptions struct {
 	*base.Options
 	Logs *logs.Options
 
-	Print  *genericclioptions.PrintFlags
-	DryRun bool
+	PrintFlags *genericclioptions.PrintFlags
+	Printer    printers.ResourcePrinter
+	DryRun     bool
 
 	// url is the argument accepted by the command. It contains the
 	// reference to where an APIService exists.
@@ -74,9 +75,9 @@ type BindOptions struct {
 // NewBindOptions returns new BindOptions.
 func NewBindOptions(streams genericclioptions.IOStreams) *BindOptions {
 	opts := &BindOptions{
-		Options: base.NewOptions(streams),
-		Logs:    logs.NewOptions(),
-		Print:   genericclioptions.NewPrintFlags("kubectl-bind"),
+		Options:    base.NewOptions(streams),
+		Logs:       logs.NewOptions(),
+		PrintFlags: genericclioptions.NewPrintFlags("kubectl-bind").WithDefaultOutput("yaml"),
 
 		Runner: func(cmd *exec.Cmd) error {
 			return cmd.Run()
@@ -92,7 +93,7 @@ func (b *BindOptions) AddCmdFlags(cmd *cobra.Command) {
 
 	b.Options.BindFlags(cmd)
 	logsv1.AddFlags(b.Logs, cmd.Flags())
-	b.Print.AddFlags(cmd)
+	b.PrintFlags.AddFlags(cmd)
 
 	cmd.Flags().BoolVar(&b.SkipKonnector, "skip-konnector", b.SkipKonnector, "Skip the deployment of the konnector")
 	cmd.Flags().BoolVarP(&b.DryRun, "dry-run", "d", b.DryRun, "If true, only print the requests that would be sent to the service provider after authentication, without actually binding.")
@@ -107,6 +108,18 @@ func (b *BindOptions) Complete(args []string) error {
 	if len(args) > 0 {
 		b.URL = args[0]
 	}
+
+	if b.DryRun {
+		b.PrintFlags.Complete("%s (dry run)")
+	}
+
+	printer, err := b.PrintFlags.ToPrinter()
+	if err != nil {
+		return err
+	}
+
+	b.Printer = printer
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


ardaguclu:
@sttts @s-urbaniak could you PTAL?. Thanks.

s-urbaniak (approved):

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -107,6 +108,18 @@ func (b *BindOptions) Complete(args []string) error {
...
+	if b.DryRun {
+		b.PrintFlags.Complete("%s (dry run)") // nolint: errcheck
+	}
+
+	printer, err := b.PrintFlags.ToPrinter()
+	if err != nil {
+		return err
+	}
+
+	b.Printer = printer
```
sttts, line 121: this is misplaced in the options IMO. Should be called further down in Run.

sttts: Or does ToPrinter include validation? Then we should call it in the validate func.

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -258,20 +264,10 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
...
-			return err
-		}
-		first := true
 		for _, request := range apiRequests {
-			printer.PrintObj(request, b.IOStreams.Out) // nolint: errcheck
-
-			// TODO: this is a hack to separate the objects. Is there anything better in the printer?
-			if !first && *b.Print.OutputFormat == "yaml" {
-				fmt.Fprintln(b.IOStreams.Out, "---") // nolint: errcheck
+			if err = b.Printer.PrintObj(request, b.IOStreams.Out); err != nil {
```
sttts, line 268: this is not good. It won't lead to a valid yaml file in case of yaml encoding, will it?

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -55,8 +55,9 @@ type BindOptions struct {
 	*base.Options
 	Logs *logs.Options
 
-	Print  *genericclioptions.PrintFlags
-	DryRun bool
+	PrintFlags *genericclioptions.PrintFlags
```
sttts, line 58: it was intentionally called Print, not PrintFlags. See how all other options here are not FooOptions, but Foo.

pkg/kubectl/bind/plugin/bind.go
ardaguclu: ToPrinter includes validation but it can not be in validate function. Because flag -> options mapping is done in complete. The cases validation and mapping intertwining mostly done in complete in kubectl.

pkg/kubectl/bind/plugin/bind.go
ardaguclu: I'll show in test cases that it is true :)

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -74,9 +75,9 @@ type BindOptions struct {
 // NewBindOptions returns new BindOptions.
 func NewBindOptions(streams genericclioptions.IOStreams) *BindOptions {
 	opts := &BindOptions{
-		Options: base.NewOptions(streams),
-		Logs:    logs.NewOptions(),
-		Print:   genericclioptions.NewPrintFlags("kubectl-bind"),
+		Options:    base.NewOptions(streams),
+		Logs:       logs.NewOptions(),
+		PrintFlags: genericclioptions.NewPrintFlags("kubectl-bind").WithDefaultOutput("yaml"),
```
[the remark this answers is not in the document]
ardaguclu: That is a good point. I added (dry run) suffix but it is not added for yaml or json cases. I agree that it does not make sense other than `yaml` and `json`.

test/e2e/framework/bind.go
```diff
@@ -54,25 +57,33 @@ func Bind(t *testing.T, authURLCh chan<- string, invocations chan<- SubCommandIn
...
 
-		return nil
+			return nil
+		}
 	}
 	err = opts.Run(ctx, authURLCh)
 	require.NoError(t, err)
+
+	if opts.DryRun {
+		_, err = yaml.YAMLToJSON(bufOut.Bytes())
```
ardaguclu, line 84: Currently this only verifies one resource yaml. I'm going to add multiple binding resource support into e2e suite but it requires a little bit effort.

test/e2e/framework/bind.go
```diff
@@ -54,25 +57,33 @@ func Bind(t *testing.T, authURLCh chan<- string, invocations chan<- SubCommandIn
...
+					Args:       cmd.Args[1:],
+					Stdin:      bs,
+				}
+			}
+			t.Logf("Running command: %s\nstdin:\n", cmd.String())
+			t.Logf("%s", bs)
 
-		return nil
+			return nil
+		}
```
sttts, line 78: why this special casing? This bind runner does not have to invoke a sub command. In that case the invocations channel will just not get any data.

pkg/kubectl/bind/plugin/bind.go
sttts: not sure I can follow.

test/e2e/framework/bind.go
```diff
@@ -54,25 +57,33 @@ func Bind(t *testing.T, authURLCh chan<- string, invocations chan<- SubCommandIn
...
 	}
 	err = opts.Run(ctx, authURLCh)
 	require.NoError(t, err)
+
+	if opts.DryRun {
+		_, err = yaml.YAMLToJSON(bufOut.Bytes())
+		require.NoError(t, err)
+	}
```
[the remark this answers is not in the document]
ardaguclu: Agree but iostreams are initialized in here. If you are ok, I can move iostreams to caller and do this check in there

test/e2e/framework/bind.go
ardaguclu: Runner is only executed if dry run is not enabled https://github.com/kube-bind/kube-bind/blob/a8214066c939e29ab648c94e4b74e62894a4fa62/pkg/kubectl/bind/plugin/bind.go#L314

in our case, test is failing when we pass `--dry-run`. Because `invocations` channel is not called by anybody(expectation is being called by runner)

ardaguclu: sorry, you are right. We just don't have to listen invocation channel and that is it.

test/e2e/framework/bind.go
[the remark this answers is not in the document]
sttts: yes, exposing them is fine.

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -55,8 +55,9 @@ type BindOptions struct {
 	*base.Options
 	Logs *logs.Options
 
-	Print  *genericclioptions.PrintFlags
-	DryRun bool
+	Print   *genericclioptions.PrintFlags
+	Printer printers.ResourcePrinter
```
sttts, line 59: nit: make is lower-case, or put a `if b.Printer == nil` around the line setting it below.

PR #145, opened 2022-12-07
Fix konnector image override bug for custom image

### LINKED ISSUE #144: --konnector-image option does not deploy correctly
If I use:

```
kubectl bind apiservice --konnector-image=... ...
```

kube-bind deploys the custom image, then immediately deploys over it with the official image. The output looks like this:

```
🚀 Deploying konnector v0.3.0 to namespace kube-bind with custom image "path/to/custom/konnector".
🚀 Deploying konnector v0.3.0 to namespace kube-bind.
```

is that intentional?

I worked around this by using `kubectl bind apiservice --skip-konnector --konnector-image=...`, which, counterintuitively, correctly deploys the konnector.

pkbhowmick:
Signed-off-by: Pulak Kanti Bhowmick <pulak@appscode.com>

Fix #144 

Only check for konnector current version when user using official/default image.

### DIFF
--- a/pkg/kubectl/bind-apiservice/plugin/konnector.go
+++ b/pkg/kubectl/bind-apiservice/plugin/konnector.go
@@ -64,10 +64,6 @@ func (b *BindAPIServiceOptions) deployKonnector(ctx context.Context, config *res
 		return err
 	}
 
-	konnectorVersion, installed, err := currentKonnectorVersion(ctx, kubeClient)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


s-urbaniak:
Thank you so much for the contribution! Do you mind to add a unit test here for this logic?

pkbhowmick:
Sure, will do that!

sttts (approved):

sttts:
I fear our testability lacking here.

PR #146, opened 2022-12-10
Fix servicebinding controller reconciler bug

### LINKED ISSUE #140: Service provider pretty name is shared
See from my talk https://youtu.be/Uv0ivz5xej4?t=1213, look at the columns. Something is wrong:

Bildschirm­foto 2022-11-21 um 19 16 30

pkbhowmick:
Signed-off-by: Pulak Kanti Bhowmick <pulak@appscode.com>

Fix #140 

As konnector is running APIServiceBinding controller for each provider cluster, so each controller should skip others provider's APIServiceBinding object.

### DIFF
--- a/pkg/konnector/controllers/cluster/servicebinding/servicebinding_reconcile.go
+++ b/pkg/konnector/controllers/cluster/servicebinding/servicebinding_reconcile.go
@@ -19,10 +19,13 @@ package servicebinding
 import (
 	"context"
 
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 
 	kubebindv1alpha1 "github.com/kube-bind/kube-bind/pkg/apis/kubebind/v1alpha1"
@@ -37,6 +40,7 @@ type reconciler struct {
 	getServiceExport  func(ns string) (*kubebindv1alpha1.APIServiceExport, error)
 	getServiceBinding func(name string) (*kubebindv1alpha1.APIServiceBinding, error)
 	getClusterBinding func(ctx context.Context) (*kubebindv1alpha1.ClusterBinding, error)
+	getSecret         func(ns, name string) (*corev1.Secret, error)
 
 	updateServiceExportStatus func(ctx context.Context, export *kubebindv1alpha1.APIServiceExport) (*kubebindv1alpha1.APIServiceExport, error)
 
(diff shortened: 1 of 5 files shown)


pkg/konnector/controllers/cluster/servicebinding/servicebinding_reconcile.go
```diff
@@ -46,7 +50,38 @@ type reconciler struct {
...
+	if !found {
+		logger.Error(err, "kubeconfig in secret does not have a current context", "namespace", ref.Namespace, "name", ref.Name)
+		return nil // nothing we can do here. The APIServiceBinding Controller will set a condition
+	}
+
+	// As konnector is running APIServiceBinding controller for each provider cluster,
+	// so each controller should skip others provider's APIServiceBinding object
+	if kubeContext.Namespace != r.providerNamespace {
+		return nil
+	}
```
sttts, line 84: wondering, is this even enough? Couldn't there be multiple connected provider clusters with the same namespace name? I would rather store the APIServiceBinding name here in the controller instance and skip other service bindings.

pkbhowmick: yes, there can be multiple clusters with the same namespace name. so it is better to keep the service binding name in controller instance. Updating...
Thanks!

pkbhowmick: Updated with service binding name.

pkg/konnector/konnector_reconcile.go
```diff
@@ -120,6 +120,7 @@ func (r *reconciler) reconcile(ctx context.Context, binding *kubebindv1alpha1.AP
 	ctrl, err := r.newClusterController(
 		binding.Spec.KubeconfigSecretRef.Namespace+"/"+binding.Spec.KubeconfigSecretRef.Name,
 		providerNamespace,
+		binding.Name,
```
sttts, line 123: it's a little bit more complicated. Compare the `ctrlContext` logic above. There The controller context contains a list of bindings it is started for. So it is also responsible for all its bindings, i.e. not only `binding.Name` here.

pkg/konnector/konnector_reconcile.go
```diff
@@ -120,6 +120,9 @@ func (r *reconciler) reconcile(ctx context.Context, binding *kubebindv1alpha1.AP
 	ctrl, err := r.newClusterController(
 		binding.Spec.KubeconfigSecretRef.Namespace+"/"+binding.Spec.KubeconfigSecretRef.Name,
 		providerNamespace,
+		func(svcBinding *kubebindv1alpha1.APIServiceBinding) bool {
+			return svcBinding.Name == binding.Name
```
sttts, line 124: nearly. You have to check that `svcBinding.Name` is in `ctrlContext.serviceBindings`, while taking the `r.lock`. Maybe it's a good idea to turn the lock into a RWMutex.

pkg/konnector/konnector_reconcile.go
```diff
@@ -120,6 +120,16 @@ func (r *reconciler) reconcile(ctx context.Context, binding *kubebindv1alpha1.AP
 	ctrl, err := r.newClusterController(
 		binding.Spec.KubeconfigSecretRef.Namespace+"/"+binding.Spec.KubeconfigSecretRef.Name,
 		providerNamespace,
+		func(svcBinding *kubebindv1alpha1.APIServiceBinding) bool {
+			r.lock.RLock()
+			defer r.lock.RUnlock()
+			// for the very first time, ctrlContext is nil
+			// and need to check for this as well
```
sttts, line 127: don't think this is necessary. The controller starts in line 146 after the `r.controllers` change.

pkg/konnector/konnector_reconcile.go
```diff
@@ -120,6 +120,16 @@ func (r *reconciler) reconcile(ctx context.Context, binding *kubebindv1alpha1.AP
...
+			if binding.Name == svcBinding.Name {
+				return true
+			}
+			return ctrlContext != nil && ctrlContext.serviceBindings.Has(svcBinding.Name)
```
sttts, line 131: This won't work here. You have to move the whole block of line 140-145 above this closure.

sttts (approved):

PR #149, opened 2023-01-30
Remove duplicate event handler from informer

matteoolivi:
The event handler on the consumer secrets informer (in the konnector's APIServiceBinding controller) was registered twice, this PR removes the second (unneeded) occurrence.

See the lines right above the removed ones to verify that the event handler is already registered.

### DIFF
--- a/pkg/konnector/controllers/servicebinding/servicebinding_controller.go
+++ b/pkg/konnector/controllers/servicebinding/servicebinding_controller.go
@@ -119,18 +119,6 @@ func NewController(
 		},
 	})
 
-	consumerSecretInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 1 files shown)


matteoolivi:
Even if it's unrelated, I'll blatantly ask here. I just noticed that the same controller that this PR touches has [a CRD Lister and Indexer](https://github.com/matteoolivi/kube-bind/blob/remove_duplicate_event_handler_registration/pkg/konnector/controllers/servicebinding/servicebinding_controller.go#L139-L140). But they're never used.

Are they there because they will be needed in the future, or by mistake?

If it's the latter I can open another PR to remove them.

sttts:
Looks good.

sttts:
about the crd lister, just a mistake. You can remove it.

sttts (approved):

PR #151, opened 2023-02-27
pkg/kubectl/bind: use native golang http server

s-urbaniak:


pkg/kubectl/bind/plugin/bind.go
```diff
@@ -181,17 +168,22 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
 			return err
 		}
 	}
+
+	auth := authenticator.NewLocalhostCallbackAuthenticator(10 * time.Minute)
+	err = auth.Start()
+	fmt.Fprintf(b.Options.ErrOut, "\n\n")
```
s-urbaniak, line 174: why did we have this in the first place? (left unchanged for the refactor)

pkg/kubectl/bind/authenticator/authenticator.go
```diff
@@ -44,113 +45,102 @@ func init() {
...
-	case <-ctx.Done():
-		return ctx.Err()
-	}
+	return nil
 }
 
-func (d *defaultAuthenticator) Endpoint(context.Context) string {
+// Endpoint returns the URL this server is listening to.
+// Start() must be called prior to this.
+func (d *LocalhostCallbackAuthenticator) Endpoint(context.Context) string {
```
sttts, line 93: why exactly the context? Is that some interface we have to follow?

pkg/kubectl/bind/plugin/bind.go
sttts: To separate it from other output I guess.

sttts (approved):

pkg/kubectl/bind/authenticator/authenticator.go
s-urbaniak: leftover, it was unsued, let me remove

PR #155, opened 2023-03-09
Claimed resources synchronization

lhaendler:
Discussion doc: https://docs.google.com/document/d/1Mfk2wLIURqAwSjoTimuZlL83BhslTiqV/edit#

# Scope
The prototype only implements one basic use case: downsync of all resources of a type.

This includes association of the resources to the api export in the backend, synchronization by the konnector, and basic prompts for the claim in the bind cli

# Current limitations
* only downsync
* Permission prompts show claim struct, not human readable
* No filtering implemented
* No support for adoption
* No support for donation
* No support for overrides, excludings, fields, alwaysRecreate


pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+}
+
+type AcceptablePermissionClaimState string
+
+const (
+	ClaimAccepted AcceptablePermissionClaimState = "Accepted"
+	ClaimRejected AcceptablePermissionClaimState = "Rejected"
+)
+
+// PermissionClaim identifies an object by GR and identity hash.
```
sttts, line 118:
```suggestion
// PermissionClaim selects objects of a GVR that the service provider ...
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+// Its purpose is to determine the added permissions that a service provider may
+// request and that a consumer may accept and allow the service provider access to.
+//
+// +kubebuilder:validation:XValidation:rule="!(self.create.donate && self.adopt)",message="donate and adopt are mutually exclusive"
+type PermissionClaim struct {
+	GroupResource `json:","`
+
+	Version string `json:"version"`
```
sttts, line 126: valdiation: required and non-empty

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// selector selects which resources are affected by this claim.
+	Selector ResourceSelector `json:"selector,omitempty"`
```
sttts, line 129: omitempty has no effect on structs. We could turn it into a pointer if that's prettier.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// required indicates whether the APIServiceBinding will work if this claim is not accepted.
+	Required bool `json:"required"`
```
sttts, line 132: default true?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// Global claims global resources for the given group/resource.
+	// This is mutually exclusive with resourceSelector.
+	// +optional
+	Global bool `json:"global,omitempty"`
```
sttts, line 137: is this `All`?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// only for owner Provider
+	//
+	// create determines whether the kube-bind konnector will sync matching objects from the
+	// provider side down to the consumer cluster.
+	Create CreateOptions `json:"create"`
```
sttts, line 143: this must be a pointer. There is no need that a service provider claims creation.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// adopt set to true means that objects created by the consumer are adopted by the provider.
+	// i.e. the provider will become the owner.
+	Adopt bool `json:"adopt"`
```
sttts, line 147: omitempty

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	// update lists a number of claimed permissions for the provider.
+	// "field" and "preserving" are mutually exclusive.
+	Update UpdateOptions `json:"update"`
+}
+
+type OnConflictOptions struct {
+	// providerOverrides will make the provider override any object that might already exist
+	// in the consumer cluster if it has the same namespaced name as a resource created by the
+	// provider.
```
sttts, line 161:
```suggestion
	// providerOverrides will make the provider override any object that might already exist
	// in the consumer cluster if it has the same namespaced name as a resource created by the
	// provider, but is not the result of syncing.
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+	// deletes the objet, the provider will recreate it. If the field is set as false,
+	// the provider will not recreate the object in case the object is deleted on the RecreateWhenConsumerSideDeleted
+	// side.
+	RecreateWhenConsumerSideDeleted bool `json:"recreateWhenConsumerSideDeleted"`
+}
+
+type CreateOptions struct {
+	// donate set to true means that a newly created object by the provider is immediately owned by hte consumer.
+	// If false, the object stays in ownership of the provider
+	Donate bool `json:"donate"`
```
sttts, line 176: omitempty

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+type OnConflictOptions struct {
+	// providerOverrides will make the provider override any object that might already exist
+	// in the consumer cluster if it has the same namespaced name as a resource created by the
+	// provider.
+	ProviderOverwrites bool `json:"providerOverrides"`
+
+	// only for owner provider
+	// When recreateWhenConsumerSideDeleted is true the provider will recreate the object
+	// in case the object is missing on the consumer side. Even if the consumer mistakenly or intentionally
```
sttts, line 166:
```suggestion
	// recreateWhenConsumerSideDeleted set to true means the provider will recreate the object
	// in case the object is missing on the consumer side. Even if the consumer mistakenly or intentionally.
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+type CreateOptions struct {
+	// donate set to true means that a newly created object by the provider is immediately owned by hte consumer.
+	// If false, the object stays in ownership of the provider
+	Donate bool `json:"donate"`
+}
+
+type UpdateOptions struct {
+	// fields are the fields owned by the owner of the claim. If the owner sets values of those
+	// fields, they will be synced to the other participant.
+	// Mutually exclusive with preservings.
```
sttts, line 182: would not use the word "owner" here. The claim owner is always the provider, i.e. claims apply to the provider only.

```suggestion
	// fields are a list of JSON Paths describing which parts of an object the provider wants to control in case 
	// the object is owned by the consumer as a whole. This fields cannot be set of provider owned objects
	// are selected in this claim.
	//
	// This field is mutually exclusive with preservings.
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+	Fields []string `json:"fields,omitempty"`
+
+	// Preservings are the fields that are preserved by the konnector during synchronization.
+	// The owner is not able to set those fields. If the owner changes the value of these fields,
+	// their change will be overwritten.
```
sttts, line 187:
```suggestion
	// preserving is a list of JSON Paths describing fields that should be preserved while updating with the 
	// object state on the provider side. This field cannot be set if consumer owned objects are selected in
	// this claim.
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+
+	//
+	// WARNING: If adding new fields, add them to the XValidation check!
+	//
+}
+
+type Owner string
+
+const Provider Owner = "Provider"
+const Consumer Owner = "Consumer"
```
sttts, line 217: go doc

pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
```diff
@@ -79,6 +79,13 @@ func (in *APIServiceExport) SetConditions(conditions conditionsapi.Conditions) {
 type APIServiceExportSpec struct {
 	APIServiceExportCRDSpec `json:",inline"`
 
+	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
+	// +kubebuilder:validation:MaxItems=2
```
sttts, line 86: why only 2?

pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
```diff
@@ -79,6 +79,13 @@ func (in *APIServiceExport) SetConditions(conditions conditionsapi.Conditions) {
 type APIServiceExportSpec struct {
 	APIServiceExportCRDSpec `json:",inline"`
 
+	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
```
sttts, line 85: don't think we are able to define a key for this list. This is rather a set.

pkg/apis/kubebind/v1alpha1/apiserviceexporttemplate_types.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2022 The Kube Bind Authors.
```
sttts, line 2: 2023

pkg/apis/kubebind/v1alpha1/apiserviceexporttemplate_types.go
```diff
@@ -0,0 +1,69 @@
...
+// are synchronized from and to the consumer cluster.
+//
+// +crd
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope=Namespaced,categories=kube-bindings
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Established",type="string",JSONPath=`.status.conditions[?(@.type=="Established")].status`,priority=5
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=`.metadata.creationTimestamp`,priority=0
+type APIServiceExportTemplate struct {
```
sttts, line 34: this is interesting, especially the selector. Don't think we have explored this side yet. Worth a discussion in the gdoc? My gut feeling is that this is really part of a demo backend, but could have different shape depending on context. For example the claims could differ depending on group membership or on input the UI.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
lhaendler: I think it's better to make it opt in. It would make people think about whether it is actually needed. Since required claims cannot be denied, we should guide users to use required as little as needed by making the default false.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
lhaendler: I mostly added it to support cluster-scoped resources. However I'm not sure if that's actually something we want to enable. The problem being that given any claim but single-name claims, the user could have access to cluster-scoped objects of other bind consumers.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
lhaendler: The owner term was taken from the selector. It describes the binding party that, if not overridden, will be the side to create and delete the resource.
My interpretation of "fields" and "preserving" was that:
"fields" is a whitelist for synchronization, meaning everything in "fields" uses the owner of the claimed resource (owner can be consumer or provider) as the source of truth. Everything not in "fields" uses the other side (if the owner is consumer it used provider) as the source of truth.

"excludes" is a blacklist, meaning that fields that are referenced in "excludes" do not use the owner of the resource as the source of truth, and everything not in "excludes" uses the owner as the source of truth.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,135 @@ type APIServiceBindingSpec struct {
...
+	// +kubebuilder:validation:Pattern="^([a-z0-9][-a-z0-9_.]*)?[a-z0-9]$"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+
+	// +kubebuilder:validation:Enum=Provider;Consumer
+	Owner Owner `json:"owner"`
+
+	//
+	// WARNING: If adding new fields, add them to the XValidation check!
```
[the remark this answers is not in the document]
lhaendler: Probably not. I would argue it makes sense to make the whole permissionClaim immutable though.
If we make it mutable, and simply take over the changes to the consumer, the provider could change the permission after it has been accepted, leading to a compromised consumer.

deploy/crd/kube-bind.io_apiservicebindings.yaml
```diff
@@ -205,9 +205,6 @@ spec:
                   - resource
                   - state
                   type: object
-                  x-kubernetes-validations:
-                  - message: donate and adopt are mutually exclusive
-                    rule: '!(self.create.donate && self.adopt)'
```
lhaendler: This validation broke the e2e tests, so i temporarily disabled it.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
s-urbaniak: can't we simply reference `GroupVersionResource`, why is `GroupResource` and `Version` separate?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
s-urbaniak: > Since required claims cannot be denied

we should make this fact part of the API doc.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
s-urbaniak: I am not convinced we should not allow cluster scoped resources. If we do, we lift the concept of a namespace to be a tenancy boundary which might actually not be desired. If used i.e. with kcp, claiming cluster-scoped resources is actually fine, as tenancy boundaries are defined one level up, at the workspace level.

pkg/apis/kubebind/v1alpha1/apiserviceexporttemplate_types.go
s-urbaniak: i am actually not understanding why this is needed.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
sttts: These claims are always from the point of view of the provider. There are two cases:
1. the provider is owner => `preserving` defines exceptions that the provider (here the owner) does not claim to own
2. the consumer is the owner => `fields` defines exceptions that the provider claims to own despite the consumer being the owner.

In other words: the subject of the claim is always the provider. There is no inversion of the subject when the owner changes.

pkg/apis/kubebind/v1alpha1/apiserviceexporttemplate_types.go
sttts: Suggestion: a different API group `example-backend.kube-bind.io/v1alpha1`

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="kubeconfigSecretRef is immutable"
 	KubeconfigSecretRef ClusterSecretKeyRef `json:"kubeconfigSecretRef"`
+
+	// permissionClaims records decisions about permission claims requested by the API service provider.
+	// Individual claims can be accepted or rejected. If accepted, the API service provider gets the
+	// requested access to the specified resources in this workspace. Access is granted per
+	// GroupResource, identity, and other properties.
```
sttts, line 94:
```suggestion
	// GroupResource and other properties like selectors.
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
sttts: `non-empty` need validation, aka minlength

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="!(has(self.autoDonate) && self.autoDonate && has(self.autoAdopt) && self.autoAdopt)",message="donate and adopt are mutually exclusive"
+type PermissionClaim struct {
+	GroupResource `json:","`
+
+	// +kubebuilder:validation:Required
+	Version string `json:"version"`
+
+	// Selector selects which resources are affected by this claim.
+	// +optional
+	Selector *ResourceSelector `json:"selector,omitempty"`
```
sttts, line 130: all the go doc must reference the lower-case field name, i.e. `selector` here. Same below

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
...
+	//
+	// +optional
+	Annotations []Matcher `json:"annotations,omitempty"`
+
+	// OverrideAnnotations is a list of claimed annotation key wildcard
+	// patterns that are synchronized from the consumer side
+	// to the provider on objects owned by the provider.
+	//
+	// +optional
+	OverrideAnnotations []Matcher `json:"overrideAnnotations,omitempty"`
```
sttts, line 199: remind me why we called this differently than for labels?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
...
+	// If false, the object stays in ownership of the provider.
+	//
+	// +optional
+	AutoDonate bool `json:"autoDonate,omitempty"`
+
+	// onConflict determines how the conflicts between objects on the consumer side
+	// will be resolved.
+	//
+	// +optional
+	OnConflict *OnConflictOptions `json:"onConflict,omitempty"`
```
sttts, line 163: this needs a default value. Otherwise, the inner defaults are not applied.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
sttts: this needs a default value in order for the inner defaults to apply

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
...
+	// Selector selects which resources are affected by this claim.
+	// +optional
+	Selector *ResourceSelector `json:"selector,omitempty"`
+
+	// Required indicates whether the APIServiceBinding will work if this claim is not accepted.
+	Required bool `json:"required"`
+
+	// +optional
+	// +kubebuilder:default={}
+	Read *ReadOptions `json:"read,omitempty"`
```
sttts, line 137: needs a better go doc.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,234 @@ type APIServiceBindingSpec struct {
...
+	AutoDonate bool `json:"autoDonate,omitempty"`
+
+	// onConflict determines how the conflicts between objects on the consumer side
+	// will be resolved.
+	//
+	// +optional
+	OnConflict *OnConflictOptions `json:"onConflict,omitempty"`
+
+	// Update lists a number of claimed permissions for the provider.
+	// "field" and "preserving" are mutually exclusive.
```
sttts, line 166: go doc needs update

contrib/deploy/crd/bootstrap.go
```diff
@@ -0,0 +1,181 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+)
+
+//go:embed *.yaml
+var raw embed.FS
+
+// CreateFromFS creates the given CRDs using the target client from the
+// provided filesystem and waits for it to become established. This call is blocking.
+func CreateFromFS(ctx context.Context, client apiextensionsv1client.CustomResourceDefinitionInterface, fs embed.FS, grs ...metav1.GroupResource) error {
```
sttts, line 46: remind me where this code comes from and why we need it? Is it from kcp? Looks familar.

contrib/example-backend/apis/examplebackend/v1alpha1/register.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+var (
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	AddToScheme   = SchemeBuilder.AddToScheme
+)
+
+const (
+	// GroupName is the group name used in this package
+	GroupName = "example.com"
```
sttts, line 32: `example-backend.kube-bind.io` would be better.

contrib/example-backend/controllers/serviceexportrequest/serviceexportrequest_controller.go
```diff
@@ -101,6 +102,7 @@ func NewController(
 			deleteServiceExportRequest: func(ctx context.Context, ns, name string) error {
 				return bindClient.KubeBindV1alpha1().APIServiceExportRequests(ns).Delete(ctx, name, metav1.DeleteOptions{})
 			},
+			crds: exporttemplate.NewCatalogue(config),
```
sttts, line 105: code follows American spelling, i.e. catalog

contrib/example-backend/exporttemplate/index.go
```diff
@@ -0,0 +1,93 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	crd "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+
+	"github.com/kube-bind/kube-bind/contrib/example-backend/apis/examplebackend/v1alpha1"
+	templates "github.com/kube-bind/kube-bind/contrib/example-backend/client/clientset/versioned"
+)
+
+type Index struct {
```
sttts, line 35: not sure I can follow this construct. What is it?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
sttts: question still holds

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+
+	// required indicates whether the APIServiceBinding will work if this claim is not accepted. If a required claim is denied, the binding is aborted.
+	Required bool `json:"required"`
+
+	// read claims read access to matching objects for the provider.
+	// reading of the claimed object(s) is always claimed.
+	// By default no labels and annotations are read. Reading of labels and annotations can be claimed
+	// optionally by adding labels and annotations items.
+	// If labels on consumer owned objects that are set by the consumer are read, labelsOnProviderOwnedObjects and
+	// annotationsOnProviderOwnedObjects can be set.
```
sttts, line 147:
```suggestion
	// read claims read access to matching objects for the provider.
	// Reading of the claimed object(s) is always claimed.
	// By default, no labels and annotations can be read by the provider. 
	// Reading of labels and annotations can be claimed in addition by specifying them explicitly.
	// If labels on consumer owned objects that are set by the consumer are read, labelsOnProviderOwnedObjects and
	// annotationsOnProviderOwnedObjects can be set.
```

Still trying to make sense of the last sentence. Might come back here during review.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+	AutoDonate bool `json:"autoDonate,omitempty"`
+
+	// onConflict determines how the conflicts between objects on the consumer cluster will be resolved.
+	//
+	// +optional
+	// +kubebuilder:default:={}
+	OnConflict *OnConflictOptions `json:"onConflict,omitempty"`
+
+	// update lists which updates to objects on the consumer cluster are claimed.
+	// By default, the whole object is synced, but metadata is not.
```
sttts, line 181:
```suggestion
	// By default, the whole object is continously synced, but metadata is not after creation.
```

Is this correct?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+
+type OnConflictOptions struct {
+	// recreateWhenConsumerSideDeleted set to true (the default) means the provider will recreate the object
+	// in case the object is missing on the consumer cluster, but has been synchronized before.
+	//
+	// If set to false, deleted provider-owned objects get deleted on the provider cluster as well.
+	//
+	// Even if the consumer mistakenly or intentionally
+	// deletes the object, the provider will recreate it. If the field is set as false,
+	// the provider will not recreate the object in case the object is deleted on the consumer cluster.
```
sttts, line 230: These are redundant? Might have been old text in our gdoc. Or is there a detail I am missing?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+	//
+	// +kubebuilder:default:=true
+	RecreateWhenConsumerSideDeleted bool `json:"recreateWhenConsumerSideDeleted"`
+}
+
+type CreateOptions struct {
+	// replaceExisting means that an existing object owned by the consumer will be replaced by the provider object.
+	//
+	// If set to false, and a conflicting consumer object exists, it is not touched.
```
sttts, line 239: nit:
```suggestion
	// If not true, and a conflicting consumer object exists, it is not touched.
```
Formally, false and unset are different values. Actually, we could also default this to `false` and remove the omitempty.

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+	//
+	// +optional
+	ReplaceExisting bool `json:"replaceExisting,omitempty"`
+}
+
+type UpdateOptions struct {
+	// fields are a list of JSON Paths describing which parts of an object the provider wants to control.
+	//
+	// This field is ignored if the owner in the claim selector is set to "Provider".
```
sttts, line 248: this wording is strange. The selector selects objects, either consumer or provider owned. This here is about the object owner, not about the selector.

pkg/apis/kubebind/v1alpha1/helpers/serviceexport.go
```diff
@@ -90,6 +90,10 @@ func CRDToServiceExport(crd *apiextensionsv1.CustomResourceDefinition) (*kubebin
 		if !crdVersion.Served {
 			continue
 		}
+		// TODO
+		if onlyFirstServingVersion && !crdVersion.Storage {
+			continue
+		}
```
sttts, line 96: would expect this to be a parameter to `CRDToServiceExport` and controlled through higher level means.

pkg/kubectl/bind-apiservice/plugin/servicebindings_test.go
```diff
@@ -0,0 +1,824 @@
+/*
```
sttts, line 1: would love to see us merge `servicebindings.go` and this test, plus the API types first, to get this in step by step instead of 6000 lines at once.

pkg/kubectl/bind/plugin/bind.go
```diff
@@ -303,6 +314,10 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
 	return nil
 }
 
+func (opts *BindOptions) Cleanup() {
+	os.Remove(opts.outFile.Name())
```
sttts, line 318: this is strange to be attached to options. Options are supposed to be just a data structure of flag values.

contrib/deploy/crd/bootstrap.go
lhaendler: I took it from https://github.com/kube-bind/kube-bind/blob/main/deploy/crd/bootstrap.go

contrib/example-backend/exporttemplate/index.go
lhaendler: The backend needs to know which CRD get which claimedresources. I stored them in APIServiceExportTemplates. `Index` lets the backend get all exported CRDs (those that are referenced by a APIServiceExportTemplate), and to look up the template for a CRD so it can attach the claimed resources

pkg/apis/kubebind/v1alpha1/helpers/serviceexport.go
lhaendler: this change can be removed. The intentions were the same as in #181

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+	// If a resources namespace matches any value in namespaces, the resource namespace is considered matching.
+	// If the claim is for a cluster-scoped resource, namespaces has to explicitly be set to an empty array to prevent defaulting to "*".
+	// If the "names" field is unset, all objects of the group/resource within the listed namespaces (or cluster) will be claimed.
+	//
+	// // +kubebuilder:validation:XValidation:rule="self.all(n, n.matches('^[A-z-]+|[*]$'))",message="only names or * are allowed"
+	// +kubebuilder:default:={"*"}
+	// +optional
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// labelSelectors is a list of label selectors matching selected resources. label selectors follow the same rules as kubernetes label selectors,
```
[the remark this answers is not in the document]
lhaendler: Disjunction for items of the array, conjunction for key-value pairs in the map

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
lhaendler: yes, let's remove them from both places

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,265 @@ type APIServiceBindingSpec struct {
...
+	// // +kubebuilder:validation:XValidation:rule="self.all(n, n.matches('^[A-z-]+|[*]$'))",message="only names or * are allowed"
+	// +kubebuilder:default:={"*"}
+	// +optional
+	Names []string `json:"names,omitempty"`
+
+	// namespaces represents namespaces where an object of the given group/resource may be managed.
+	// Namespaces matches against the metadata.namespace field. A value of "*" matches namespaced objects across all namespaces.
+	// If a resources namespace matches any value in namespaces, the resource namespace is considered matching.
+	// If the claim is for a cluster-scoped resource, namespaces has to explicitly be set to an empty array to prevent defaulting to "*".
+	// If the "names" field is unset, all objects of the group/resource within the listed namespaces (or cluster) will be claimed.
```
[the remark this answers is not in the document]
lhaendler: What should we do for cluster-scoped objects?

pkg/konnector/controllers/cluster/claimedresources/claimedresources_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog/v2"
+
+	kubebindv1alpha1 "github.com/kube-bind/kube-bind/pkg/apis/kubebind/v1alpha1"
+)
+
+const annotation = "kube-bind.io/resource-owner"
```
sttts, line 33: `kube-bind.io/owner` would be better

mjudeikis:
implemented as part of https://github.com/kube-bind/kube-bind/pull/304

PR #156, opened 2023-03-27
pkg/kubectl: remove unused urlCh

s-urbaniak:
The `urlCh` is a remainder of a previous implementation. It is currently unused, hence let's remove it. Passing channels over so many layers is probably not a good idea in the first place.

/cc @sttts


s-urbaniak:
Oh i see, we're using it in e2e tests

PR #159, opened 2023-03-31
Add konnector-image flag for top bind command

### LINKED ISSUE #154: Add mechanism to use custom konnector image
**Current scenario:**
Currently, in kube-bind there is no mechanism to use custom image for konnector. 

**Expected scenario:**
1. User can skip default konnector and choose provider konnector(should be open source so that user can trust the provider's konnector) via flags while running kubectl bind.
2. Provider(example-backend) can send the konnector image in the bind process(possibly with the bind response alongside with the provider kubeconfig)

pkbhowmick:
Fix #154

### DIFF
--- a/pkg/kubectl/bind/plugin/bind.go
+++ b/pkg/kubectl/bind/plugin/bind.go
@@ -275,7 +275,6 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
 			"apiservice",
 			"--remote-kubeconfig-namespace", secret.Namespace,
 			"--remote-kubeconfig-name", secret.Name,
-			"--konnector-image", b.KonnectorImageOverride,
 			"-f", "-",
 		}
 		b.flags.VisitAll(func(flag *pflag.Flag) {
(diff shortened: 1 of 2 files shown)


pkg/kubectl/bind/plugin/bind.go
```diff
@@ -279,6 +283,10 @@ func (b *BindOptions) Run(ctx context.Context, urlCh chan<- string) error {
 			}
 		})
 
+		if b.KonnectorImageOverride != "" {
+			args = append(args, "--konnector-image"+b.KonnectorImageOverride)
```
sttts, line 287: not `"--konnector-image="+b.KonnectorImageOverride` ?

pkbhowmick: Not actually. We are putting the flag key and value in a string slice which is handled according. So no need for '='.

pkbhowmick: Sorry, misunderstood before. Fixed it now.

sttts (approved):

PR #175, opened 2023-07-27
Add supports to cluster-scoped APIs

waltforme:
This PR tries to do two things regarding cluster-scoped APIs.
1. Completely enable the exporting and syncing of cluster-scoped APIs. As mentioned by @sttts in the [slack channel](https://kubernetes.slack.com/archives/C046PRXNJ4W/p1689088717859549?thread_ts=1688665656.221059&cid=C046PRXNJ4W), the exporting of cluster-scoped APIs is almost working. Some small tweaks are added in 934bda0 to make it complete.
2. Resolve the name conflicts for cluster-scoped API objects. Most work of the PR is for this second thing. To do that, a shim is implemented to a) translate between consumer side object names and provider side object names, b) filter object names for individual consumers. I came up with a diagram to make it easier to be reviewed. The shim is represented by the black curve in the middle of the diagram (feel free to click to zoom-in or visit the [original drawing](https://docs.google.com/drawings/d/1DH609qJW8uAuowmMx2Bo957edRJ4Mv0_BzLPmLbiX6U/)):

image

This PR doesn't resolve the permission question, as discussed [in this thread](https://kubernetes.slack.com/archives/C046PRXNJ4W/p1689088050065169?thread_ts=1688665656.221059&cid=C046PRXNJ4W). But the point is well taken. Looks like something like node authorizer is a way to go, and I'm happy to continue working on that.

### DIFF
--- a/contrib/example-backend/deploy/01-clusterrole.yaml
+++ b/contrib/example-backend/deploy/01-clusterrole.yaml
@@ -11,6 +11,11 @@ rules:
 - apiGroups:
     - ""
   resources:
+    - "namespaces"
+  verbs: ["get"]
+- apiGroups:
+    - ""
+  resources:
     - "secrets"
   verbs: ["get", "watch", "list"]
 - apiGroups:
--- a/pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
+++ b/pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
@@ -121,26 +121,35 @@ func ClearClusterNs(obj *unstructured.Unstructured, clusterNs string) error {
 	return nil
 }
 
-func TranslateFromDownstream(obj *unstructured.Unstructured, clusterNs, clusterNsUID string) (*unstructured.Unstructured, error) {
-	err := InjectClusterNs(obj, clusterNs, clusterNsUID)
+// TranslateFromDownstream mutates a cluster-scoped object in place by injecting the cluster namespace
+// and switching to its corresponding upstream name
+func TranslateFromDownstream(obj *unstructured.Unstructured, clusterNs, clusterNsUID string) error {
+	copy := obj.DeepCopy()
+	err := InjectClusterNs(copy, clusterNs, clusterNsUID)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	SwitchToUpstreamName(obj, clusterNs)
-	return obj, nil
+	SwitchToUpstreamName(copy, clusterNs)
+	*obj = *copy
+	return nil
 }
 
-func TranslateFromUpstream(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+// TranslateFromUpstream mutates a cluster-scoped object in place by clearing the injected cluster namespace
+// and switching to its corresponding downstream name
+func TranslateFromUpstream(obj *unstructured.Unstructured) error {
 	clusterNs, err := ExtractClusterNs(obj)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	err = ClearClusterNs(obj, clusterNs)
+
+	copy := obj.DeepCopy()
+	err = ClearClusterNs(copy, clusterNs)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	SwitchToDownstreamName(obj, clusterNs)
-	return obj, nil
+	SwitchToDownstreamName(copy, clusterNs)
+	*obj = *copy
+	return nil
 }
 
 func findOwnerReferenceToClusterNs(ors []metav1.OwnerReference, clusterNs string) (int, bool) {
--- a/pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
+++ b/pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2023 The Kube Bind Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterscoped
+
+import (
+	"errors"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ClusterNsAnnotationKey is the annotation key to identify the cluster namespace that corresponds to
+// provider side copy of a cluster-scoped object. Its value is the corresponding cluster namespace name.
+const ClusterNsAnnotationKey = "kube-bind.io/cluster-namespace"
+
+// Prepend adds clusterNs and a dash before name.
+func Prepend(name, clusterNs string) string {
+	return clusterNs + "-" + name
+}
+
+// Behead tries to remove clusterNs and a dash that goes before name.
+// If name doesn't start with clusterNs, name is returned unchanged.
+func Behead(name, clusterNs string) string {
+	return strings.TrimPrefix(name, clusterNs+"-")
+}
+
+// SwitchToUpstreamName switches the name of a downstream
+// cluster-scoped object by prepending the cluster namespace name.
+func SwitchToUpstreamName(obj *unstructured.Unstructured, clusterNs string) {
+	downstreamName := obj.GetName()
+	upstreamName := Prepend(downstreamName, clusterNs)
+	obj.SetName(upstreamName)
+}
+
+// SwitchToDownstreamName switches the name of a upstream
+// cluster-scoped object by removing the cluster namespace name as the prefix.
+func SwitchToDownstreamName(obj *unstructured.Unstructured, clusterNs string) {
+	upstreamName := obj.GetName()
+	downstreamName := Behead(upstreamName, clusterNs)
+	obj.SetName(downstreamName)
+}
+
+// InjectClusterNs injects the given cluster namespace (1) as an annotation,
+// and (2) as a owner reference to the cluster namespace.
+func InjectClusterNs(obj *unstructured.Unstructured, clusterNs, clusterNsUID string) error {
+	ans := obj.GetAnnotations()
+	existing, foundAn := ans[ClusterNsAnnotationKey]
+	if foundAn && existing != clusterNs {
+		return errors.New("mismatch between existing cluster namespace and given cluster namespace")
+	}
+
+	ors := obj.GetOwnerReferences()
+	idx, foundOr := findOwnerReferenceToClusterNs(ors, clusterNs)
+	if foundOr && ors[idx].Name != clusterNs {
+		return errors.New("mismatch between existing cluster namespace and given cluster namespace")
+	}
+
+	if !foundAn {
+		if ans == nil {
+			ans = map[string]string{}
+		}
+		ans[ClusterNsAnnotationKey] = clusterNs
+		obj.SetAnnotations(ans)
+	}
+
+	if !foundOr {
+		ors = append(ors, metav1.OwnerReference{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+			Name:       clusterNs,
+			UID:        types.UID(clusterNsUID),
+		})
+		obj.SetOwnerReferences(ors)
+	}
+
+	return nil
+}
+
+// ExtractClusterNs extracts the corresponding cluster namespace name
+// from a cluster-scoped object by reading the annotation.
+func ExtractClusterNs(obj *unstructured.Unstructured) (string, error) {
+	ans := obj.GetAnnotations()
+	clusterNs, ok := ans[ClusterNsAnnotationKey]
+	if !ok {
+		return "", errors.New("cluster namespace annotation not found")
+	}
+	return clusterNs, nil
+}
+
+// ClearClusterNs clears the given cluster namespace in a cluster-scoped object,
+// including both the annotation and the owner reference.
+func ClearClusterNs(obj *unstructured.Unstructured, clusterNs string) error {
+	ans := obj.GetAnnotations()
+	delete(ans, ClusterNsAnnotationKey)
+	obj.SetAnnotations(ans)
+
+	ors := obj.GetOwnerReferences()
+	idx, foundOr := findOwnerReferenceToClusterNs(ors, clusterNs)
+	if foundOr {
+		ors[idx] = ors[len(ors)-1]
+		ors = ors[:len(ors)-1]
+		obj.SetOwnerReferences(ors)
+	}
+
+	return nil
+}
+
+func TranslateFromDownstream(obj *unstructured.Unstructured, clusterNs, clusterNsUID string) (*unstructured.Unstructured, error) {
+	err := InjectClusterNs(obj, clusterNs, clusterNsUID)
+	if err != nil {
+		return nil, err
+	}
+	SwitchToUpstreamName(obj, clusterNs)
+	return obj, nil
+}
+
+func TranslateFromUpstream(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	clusterNs, err := ExtractClusterNs(obj)
+	if err != nil {
+		return nil, err
+	}
+	err = ClearClusterNs(obj, clusterNs)
+	if err != nil {
+		return nil, err
+	}
+	SwitchToDownstreamName(obj, clusterNs)
+	return obj, nil
+}
+
+func findOwnerReferenceToClusterNs(ors []metav1.OwnerReference, clusterNs string) (int, bool) {
+	if ors == nil {
+		return -1, false
+	}
+	for i, or := range ors {
+		if or.Kind == "Namespace" && or.Name == clusterNs {
+			return i, true
+		}
+	}
+	return -1, false
+}
--- a/pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
+++ b/pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtimeschema "k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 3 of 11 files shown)


pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -0,0 +1,123 @@
...
+	return clusterNs + "-" + name
+}
+
+func Behead(name, clusterNs string) string {
+	return strings.TrimPrefix(name, clusterNs+"-")
+}
+
+// SwitchToUpstreamName switches the name of a downstream
+// cluster-scoped object by prepending the cluster namespace name
+func SwitchToUpstreamName(obj *unstructured.Unstructured, clusterNs string) {
```
sttts, line 38:
```suggestion
func ToUpstreamName(obj *unstructured.Unstructured, clusterNs string) {
```

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if ors[0].Kind != "Namespace" {
+		return errors.New("object owned by something other than a namespace")
+	}
+	ors = ors[:0]
+	obj.SetOwnerReferences(ors)
+	return nil
+}
+
+func TranslateFromDownstream(obj *unstructured.Unstructured, clusterNs, clusterNsUID string) (*unstructured.Unstructured, error) {
```
sttts, line 103: this mutates. Intentional? The signature does not look like that. Would either rename to SomethingInPlace or better remove the unstructured return value. Same below.

pkg/konnector/controllers/cluster/serviceexport/spec/spec_controller.go
```diff
@@ -104,25 +106,62 @@ func NewController(
...
+				} else {
+					obj, err := clusterscoped.TranslateFromDownstream(obj, providerNamespace, providerNamespaceUID)
+					if err != nil {
+						return nil, err
+					}
+					if created, err := providerClient.Resource(gvr).Namespace(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{}); err != nil {
+						return nil, err
+					} else {
+						return clusterscoped.TranslateFromUpstream(created)
+					}
```
sttts, line 136:
```suggestion
					created, err := providerClient.Resource(gvr).Namespace(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{})
					if err != nil {
						return nil, err
					}
					return clusterscoped.TranslateFromUpstream(created)
```

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -0,0 +1,123 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func Prepend(name, clusterNs string) string {
+	return clusterNs + "-" + name
+}
+
+func Behead(name, clusterNs string) string {
```
sttts, line 32: is behead a word? Why not Strip or Trim?

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// ExtractClusterNs extracts the corresponding cluster namespace name
+// from a cluster-scoped object.
+func ExtractClusterNs(obj *unstructured.Unstructured) (string, error) {
+	ors := obj.GetOwnerReferences()
+	if len(ors) == 0 {
+		return "", errors.New("owner references not found in object")
+	}
+	if len(ors) > 1 {
```
sttts, line 76: wondering whether this is a problematic constraint. There can be other owners IMO.

pkg/konnector/controllers/cluster/serviceexport/spec/spec_controller.go
```diff
@@ -104,25 +106,62 @@ func NewController(
...
+					obj := got.(*unstructured.Unstructured)
+					return clusterscoped.TranslateFromUpstream(obj.DeepCopy())
 				}
-				return obj.(*unstructured.Unstructured), nil
 			},
 			createProviderObject: func(ctx context.Context, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
-				return providerClient.Resource(gvr).Namespace(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{})
+				if ns := obj.GetNamespace(); ns != "" {
+					return providerClient.Resource(gvr).Namespace(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{})
+				} else {
```
sttts, line 127: please avoid unidiomatic `if ... { ... return } else { ... }`

contrib/example-backend/deploy/01-clusterrole.yaml
```diff
@@ -8,6 +8,11 @@ rules:
   resources:
   - "apiserviceexportrequests"
   verbs: ["create","delete","patch","update","get","list","watch"]
+- apiGroups:
+    - ""
+  resources:
+    - "namespaces"
+  verbs: ["get"]
```
sttts, line 15: where is this needed?

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
sttts: also go doc please

sttts:
Instead of a node authorizer like construct, we could also replicate the mechanism we already have for namespaces, i.e. `APIServiceNamespace` with an RBAC controller in the backend. Wouldn't that be a much simpler approach? A node authorizer is much deeper in the system, i.e. a webhook or compiled in. A controller wouldn't need that.

sttts:
What I cannot completely follow is how the multins informer logic is working with cluster-scoped resources. Can you elaborate?

sttts:
Last but not least, we really want e2e tests for this.

MikeSpreitzer:
While the authorization is a real concern, I think we agreed that it is separable.  The use cases that are most urgent for me need this naming fix but can live without the authorization fix for a while (due to their own point-in-time limitations; eventually they will need the authorization fix too, but it does not need to be part of this PR).

waltforme:
Force-push to https://github.com/kube-bind/kube-bind/commit/151ece8034ce4e8754184a9f50a9cc68e5e3c584 is done after rebasing onto latest main and squashing unnecessary commits.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: Thank you. Agreed and removed the unstructured return values in 94cf53e.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: > also go doc please

Done in d9cdc0f.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: Indeed. I fixed this mainly in 7530710.

pkg/konnector/controllers/cluster/serviceexport/spec/spec_controller.go
waltforme: Thank you. Fixed this throughout this PR.

waltforme:
> What I cannot completely follow is how the multins informer logic is working with cluster-scoped resources. Can you elaborate?

Sure. I don't know whether I understand the question correctly but let me try to elaborate.

I saw the consumer side (konnector's) service export controller checks the scope here https://github.com/kube-bind/kube-bind/blob/ddf9f11835b611082c50d0adaa0d5b8d332fbe99/pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go#L166 before creating the providerInf for both the spec and status controllers. I also saw that, if the checked scope is "Cluster", a plain `dynamicinformer.DynamicSharedInformerFactory`, rather than a homemade `multinsinformer.DynamicMultiNamespaceInformer`, is used, in `GetterInformerWrapper.Delegate`. This plain DynamicSharedInformerFactory works for cluster-scoped objects.

pkg/konnector/controllers/cluster/serviceexport/spec/spec_controller.go
waltforme: Thank you. I make a commit for the suggested changes.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: > is behead a word? Why not Strip or Trim?

Yes! `Behead` is a word. I wanted to be as accurate as possible to catch the meaning of 'trim from the head but not tail', just like the function name `Prepend` catches similar meaning. So I looked into dictionaries and (for my first time) learnt this word, which is the closest one I can find.

waltforme:
> Instead of a node authorizer like construct, we could also replicate the mechanism we already have for namespaces, i.e. `APIServiceNamespace` with an RBAC controller in the backend. Wouldn't that be a much simpler approach? A node authorizer is much deeper in the system, i.e. a webhook or compiled in. A controller wouldn't need that.

Thanks for the ideas. I'll think about them and discuss with/consult you. Could we live without the authorization fix for a while (as @MikeSpreitzer suggests) and handle the authorization in other PRs?

waltforme:
> Last but not least, we really want e2e tests for this.

Indeed. I added e2e tests for cluster-scoped APIs in e296a60.

waltforme:
Force-push to 5cf1e1d is done after reorganizing commits and squashing unnecessary ones.

Note that the changes to styles and docs are consolidated to the last commit, so styles are not perfect until this last one.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -29,10 +29,13 @@ import (
 // a cluster-scoped object. Its value is the corresponding cluster namespace name.
 const ClusterNsAnnotationKey = "kube-bind.io/cluster-namespace"
 
+// Prepend adds clusterNs and a dash before name.
 func Prepend(name, clusterNs string) string {
 	return clusterNs + "-" + name
 }
 
+// Behead tries to remove clusterNs and a dash that goes before name.
+// If name doesn't start with clusterNs, name is returned unchanged.
```
sttts, line 38: nit: sqaush this kind of commits into the commit where these functions were added. Makes life of the reviewer much easier.

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
```diff
@@ -0,0 +1,153 @@
...
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ClusterNsAnnotationKey is the annotation key to identify the cluster namespace that corresponds to
+// a cluster-scoped object. Its value is the corresponding cluster namespace name.
+const ClusterNsAnnotationKey = "kube-bind.io/cluster-namespace"
```
sttts, line 30: be clear where this is used. On the client-side objects?

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
sttts: in a later commit? Please squash the later change here.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -160,6 +161,18 @@ func (r *reconciler) ensureControllers(ctx context.Context, name string, export
 
 	dynamicConsumerClient := dynamicclient.NewForConfigOrDie(r.consumerConfig)
 	dynamicProviderClient := dynamicclient.NewForConfigOrDie(r.providerConfig)
+
+	providerNamespaceUID := ""
```
sttts, line 165: Still trying to wrap my head around this UID. It's used as owner ref eventually on upsynced objects? Wondering about security implications: the provider does not trust the konnector. Does it have to trust that the right owner-ref is set?

sttts:
Code-wise this looks very good. I think it's a good first step.

On the higher level, here are some questions (not a blocker to move forward if there is will to address them in follow-ups):

1. being able to create objects on the provider side requires trust into the konnector. This trust does not exist. I.e. we need a way to authorize the creation of certain kind of objects, e.g. with the right prefix and right owner-ref.
2. there are different ways to map cluster-scoped objects in the consumer scope to the provider. This PR only follows one: `<namespace>-` prefix indicating it belongs to the binding in that namespace. I can see at least these models:
   1. `<namespace>-<downstream-name>`  as implemented
   2. `<downstream-name>` without any mapping, e.g. via a kcp backend where the consumer owns the complete provider cluster aka workspace
   3. `<namespace>/<downstream-name>` where the cluster-scoped CRD downstream becomes namespace-scoped upstream.
   I think we should have a way to specify that in the `APIServiceExport`, similarly to the informer scope we already have https://github.com/kube-bind/kube-bind/blob/main/pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go#L92.
3. about `<namespace>-<downstream-name>` as implemented, a variant would be that on create the konnector has to "request" access (via RBAC) to the cluster-scoped object, e.g. through another objects in `<namespace>`. But then we are already near to the third variant above. In other words, the first (implemented) variant does only really make sense if you want isolation via naming, but no isolation via RBAC on creation. This is a pretty weak security model, but maybe it has its uses.

What I would propose as follow-up is to add an API field to the export type like `clusterScopeIsolation` with one alternative `prefixed` as implemented here in the PR. We can add `none` and `namespaced` later.

sttts (approved):

test/e2e/bind/happy-case_test.go
```diff
@@ -69,172 +70,280 @@ func testHappyCase(t *testing.T, scope kubebindv1alpha1.Scope) {
...
 	framework.StartKonnector(t, consumerConfig, "--kubeconfig="+consumerKubeconfig)
 
-	consumerClient := framework.DynamicClient(t, consumerConfig).Resource(
-		schema.GroupVersionResource{Group: "mangodb.com", Version: "v1alpha1", Resource: "mangodbs"},
-	).Namespace("default")
-	providerClient := framework.DynamicClient(t, providerConfig).Resource(
-		schema.GroupVersionResource{Group: "mangodb.com", Version: "v1alpha1", Resource: "mangodbs"},
-	)
+	serviceGVR := schema.GroupVersionResource{Group: "mangodb.com", Version: "v1alpha1", Resource: "mangodbs"}
+	if scope == kubebindv1alpha1.ClusterScope {
```
sttts, line 74: I think we have two scopes now:
1. namespaced resource, but with cluster or namespace scoped informers.
2. cluster scoped resources, always with cluster scoped informers.

sttts: I.e. we want to restore this test to verify both modes in (1).

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: Done in latest push.

test/e2e/bind/happy-case_test.go
waltforme: > I.e. we want to restore this test to verify both modes in (1).

Done in latest push, and also squashed into the commit that generalizes the e2e test.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
waltforme: Yes, it is on upsynced objects. For example, in the `root:provider` kcp workspace, the Foo `test` refers to its cluster namespace `kube-bind-flbkk`.
```
👉 kube-bind $ kubectl ws .
Current workspace is "root:provider".
👉 kube-bind $ kubectl get ns
NAME                      STATUS   AGE
default                   Active   4d20h
kube-bind-flbkk           Active   4d19h
kube-bind-flbkk-default   Active   70m
👉 kube-bind $ kubectl get foo
NAME                   AGE
kube-bind-flbkk-test   70m
👉 kube-bind $ kubectl get foo kube-bind-flbkk-test -oyaml | yq .metadata.ownerReferences
- apiVersion: v1
  kind: Namespace
  name: kube-bind-flbkk
  uid: 3cb1bc1d-dfd0-4121-9801-37abe86f7d87
👉 kube-bind $ kubectl get ns kube-bind-flbkk -oyaml | yq .metadata.uid
3cb1bc1d-dfd0-4121-9801-37abe86f7d87
```

For the security/trust aspect (as listed in item 1 in the [comment below](https://github.com/kube-bind/kube-bind/pull/175#issuecomment-1721074910)), maybe I'm still not following why 'This trust does not exist'. The konnector holds a kubeconfig that is sent from the provider with proper access, no?

pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go
waltforme: Clarified this, also squashed into the commit that implements the shim.

test/e2e/bind/happy-case_test.go
```diff
@@ -36,205 +36,318 @@ import (
...
 
-	testHappyCase(t, kubebindv1alpha1.NamespacedScope)
+	// namespaced resource, with namespace scoped informers
+	testHappyCase(t, kubebindv1alpha1.NamespacedScope, kubebindv1alpha1.NamespacedScope)
+	// namespaced resource, but with cluster scoped informers
+	testHappyCase(t, kubebindv1alpha1.NamespacedScope, kubebindv1alpha1.ClusterScope)
 }
 
-func testHappyCase(t *testing.T, scope kubebindv1alpha1.Scope) {
+func testHappyCase(t *testing.T, resourceScope, informerScope kubebindv1alpha1.Scope) {
```
sttts, line 60: nit (follow-up is fine): would use `ResourceScope` from apiextension-apiserver for the first https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types.go#L255

waltforme:
Force-push to https://github.com/kube-bind/kube-bind/commit/79bc00e7cc4d892e22d557d9c7eae4cd8d02ae73 was done after responding to recent comments and squashing changes to appropriate commits.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
sttts: This is connected to [my comment of different isolation models](https://github.com/kube-bind/kube-bind/pull/175#issuecomment-1721074910) we might want to implement. This "prefixed" mode is by design not well isolated because you need arbitrary create access and watch permissions. We will want other modes which give more security, especially for Kubernetes providers.

When I say "trust into konnector does not exist" I mean that the provider cannot rely on the consumer running the official konnector. It could be a modified versions that tries malicious requests on the provider cluster.

test/e2e/bind/happy-case_test.go
waltforme: Thanks for the suggestion! Just did this in #180.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
waltforme: I think I had a better understanding now --- we were leveraging the isolation that offered by k8s namespaces, to limit the access of the potentially malicious konnectors.

I guess this answered one of my dumb questions why there are namespaced `APIServiceNamespace` living in the cluster namespaces.

Is the 'multinsinformer' also driven by this security consideration (we want a consumer be only informed about objects in a few namespaces)?

PR #185, opened 2023-10-18
Claimed resources api

lhaendler:
based on discussions from https://github.com/kube-bind/kube-bind/pull/155

### DIFF
--- a/pkg/kubectl/bind-apiservice/plugin/servicebindings_test.go
+++ b/pkg/kubectl/bind-apiservice/plugin/servicebindings_test.go
@@ -26,7 +26,7 @@ import (
 	kubebindv1alpha1 "github.com/kube-bind/kube-bind/pkg/apis/kubebind/v1alpha1"
 )
 
-func TestHumanReadablePromt(t *testing.T) {
+func TestHumanReadablePrompt(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
@@ -160,7 +160,7 @@ func TestHumanReadablePromt(t *testing.T) {
 					ObjectSelector: &kubebindv1alpha1.ObjectSelector{
 						Owner: kubebindv1alpha1.Provider,
 					},
-					OwnerTransfer: kubebindv1alpha1.OwnerTransferDonate,
+					OwnerTransfer: kubebindv1alpha1.OwnerTransferNone,
 				},
 				Required: true,
 			},


pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,261 @@ type APIServiceBindingSpec struct {
...
+// acceptablePermissionClaim is a permission claim that stores the users acceptance in the field state. Only accepted permission claims are reconciled.
+type AcceptablePermissionClaim struct {
+	PermissionClaim `json:",inline"`
+
+	// state indicates if the claim is accepted or rejected.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=Accepted;Rejected
+	State AcceptablePermissionClaimState `json:"state"`
```
sttts, line 109: how is the flow? The cli (or some other means under the user identity) creates these acceptable claims and sets the accepted or rejected? What is the consequence of rejected (vs. not being listed)?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
```diff
@@ -87,8 +87,261 @@ type APIServiceBindingSpec struct {
...
+	// patterns that are synchronized from the consumer cluster
+	// to the provider on objects owned by the provider.
+	//
+	// +optional
+	OverrideAnnotations []Matcher `json:"overrideAnnotations,omitempty"`
+}
+
+type Matcher struct {
+	// +optional
+	Pattern string `json:"pattern,omitempty"`
```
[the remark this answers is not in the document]
sttts: Also we should make this required for now, until we have another type and can make this a sum type.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+type PermissionClaimOnConflictOptions struct {
+	// recreateWhenConsumerSideDeleted set to true (the default) means the provider will recreate the object
+	// in case the object is missing on the consumer cluster, but has been synchronized before.
+	//
+	// If set to false, deleted provider-owned objects get deleted on the provider cluster as well.
+	//
+	// +kubebuilder:default:=true
+	RecreateWhenConsumerSideDeleted bool `json:"recreateWhenConsumerSideDeleted"`
```
mjudeikis, line 154: These looks like they will be mutual exlusive too, maybe something similar to:
```
// Similarly to above, these are constants to support HTTP PATCH utilized by
// both the client and server that didn't make sense for a whole package to be
// dedicated to.
type PatchType string

const (
	JSONPatchType           PatchType = "application/json-patch+json"
	MergePatchType          PatchType = "application/merge-patch+json"
	StrategicMergePatchType PatchType = "application/strategic-merge-patch+json"
	ApplyPatchType          PatchType = "application/apply-patch+yaml"
)
```
As I bet these will be more here, like `ForceRecreate`, `DeleteAndRecreate`, etc

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+type PermissionClaimCreateOptions struct {
+	// replaceExisting means that an existing object owned by the consumer will
+	// be replaced by the provider object.
+	//
+	// If not true, and a conflicting consumer object exists, it is not touched.
+	//
+	// +optional
+	ReplaceExisting bool `json:"replaceExisting,omitempty"`
```
mjudeikis, line 164: Same as above. We could reuse these if done right.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+	//
+	// +optional
+	AutoDonate bool `json:"autoDonate,omitempty"`
+
+	// onConflict determines how conflicts between objects on the consumer
+	// and provider clusters will be resolved.
+	//
+	// +optional
+	// +kubebuilder:default:={}
+	OnConflict *PermissionClaimOnConflictOptions `json:"onConflict,omitempty"`
```
mjudeikis, line 97: Somehow this does not look right. You will have conflict on Create/Update only. Now this still being optional but implied to get and Deletes too. Some logical shuffling might be needed. Like Clients logic: `Delete(ctx context.Context, name string, opts v1.DeleteOptions) error` where DeleteOptions are only for delete struct and we can reuse them inside.

mjudeikis (changes requested):
Thanks for this! APIS are hard :/ 
I think we could try modeling this more like client-go. Its not ideal, but is what people are used to. I mean like 

`Create(ctx context.Context, aPIBinding *v1alpha1.APIBinding, opts v1.CreateOptions) (*v1alpha1.APIBinding, error)`
 Try to isolate things into duplets - `Action(Get,Update,Create,Delete)` and `ActionsOptions` so stucts would be something like:
```
type PermissionClaim struct {
	GroupResource `json:","
	PermissionClaimActions `json:","
	...
}

type PermissionClaimActions{
  Get *PermissionClaimGetOptions `json:"get,omitempty"`
  Create *PermissionClaimCreateOptions `json:"create,omitempty"`
```
and thing like `OnConflict *PermissionClaimOnConflictOptions `json:"onConflict,omitempty"`` is explisit to Update/Create only and nested inside options, not on the parent struct.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.autoDonate) && self.autoDonate && has(self.autoAdopt) && self.autoAdopt)",message="donate and adopt are mutually exclusive"
+type PermissionClaim struct {
+	GroupResource `json:","`
+
+	// version is the version of the claimed resource.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength:=1
+	Version string `json:"version"`
```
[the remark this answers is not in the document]
sttts: The version also determines what to sync. So yes, we have to know.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+	AutoAdopt bool `json:"autoAdopt,omitempty"`
+
+	// autoDonate set to true means that a newly created object synced from the
+	// provider to the consumer cluster is immediately donated to the consumer
+	// by marking it as owned by the consumer on both sides.
+	//
+	// autoDonate is mutually exclusive with autoAdopt.
+	//
+	// +optional
+	AutoDonate bool `json:"autoDonate,omitempty"`
```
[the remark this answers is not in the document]
sttts: ownerTransfer = Adopt|Donate|None ?

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+	// +kubebuilder:default:={}
+	OnConflict *PermissionClaimOnConflictOptions `json:"onConflict,omitempty"`
+
+	// update lists which updates to objects on the consumer cluster are claimed.
+	// By default, the whole object is synced, but metadata is not.
+	//
+	// Note that update permissions do not imply create permissions.
+	//
+	// +optional
+	Update *PermissionClaimUpdateOptions `json:"update,omitempty"`
```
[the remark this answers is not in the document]
sttts: it's not really about requests. Hence not "get".

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
sttts: you write above to move this into update or create. Neither is really a fit.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
sttts: what is "above"?

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
sttts: Maybe `create.onDelete: Recretate|DeleteOnProviderSide`.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
```diff
@@ -0,0 +1,273 @@
...
+	// +kubebuilder:validation:MinLength:=1
+	Version string `json:"version"`
+
+	// selector restricts which objects of the given resource are being claimed.
+	// If unset, all objects across all namespaces are being claimed, both
+	// consumer- and provider owned.
+	//
+	// +optional
+	// +kubebuilder:default:={}
+	Selector *ResourceSelector `json:"selector,omitempty"`
```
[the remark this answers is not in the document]
sttts: it's actually `objectSelector`. We select objects, not resources. Resources are pods, namespaces, services, as in GVRs.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
mjudeikis: I mean, can we have more of these and be exclusive? Like: 
```
CreateStrategy:
    DeleteAndCreate (some object in k8s world cant be updated)
    UpdateInplace
```

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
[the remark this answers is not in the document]
lhaendler: Why not '*' itself?

pkg/apis/kubebind/v1alpha1/apiservicebinding_types.go
lhaendler: rejected claims are not reconciled. Saving them as rejected could allow the user to accept them later on via cli.

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
Danil-Grigorev: I’m not sure if it was brought up in the previous discussions, but what about deletion scenarios? Currently an object removed from the consumer cluster is triggering deletion on the provider side. If you remove the object on the provider side, it will be re-created from the consumer copy. Meaning that in the `Recreate` state there is no way to remove the object from the cluster without stopping the konnector instance. Shouldn’t it be more general? `Recreate|FollowDeletion` allowing for both sides being equal in having right for resource eviction?

lhaendler: That's where the `onConflict` part of the field name comes from. The action is performed when the lifecycle of the object conflicts with the state on the cluster.

> Meaning that in the Recreate state there is no way to remove the object from the cluster without stopping the konnector instance
For provider owned objects, that is true (short of undoing the binding). Claimed resources are objects that support objects of a bound api. For example, creating a `MangoDB` object should spawn a secret containing credentials to use the MangoDB instance. It is expected that the supporting objects are removed upon deletion of the 'main' resource.

The behavior in case of deletion looks like the following:

## Recreate: false

**provider** owned object deleted: not recreated on the consumer side. Since the provide still manages the object, the provider will eventually delete it on the provider cluster.

**consumer** owned object: the object will be deleted on the provider cluster, since the consumer object controls the lifecycle

## Recreate: true

**provider** owned object: the object is created again on the consumer cluster, once the provider deletes it on the provider cluster, it is deleted on the consumer cluster

**consumer** owned object: the object will be deleted on the provider cluster, since the consumer object controls the lifecycle

lhaendler: maybe onConflict.userDeletesProviderOwnedObject = Recreate | Orphan?

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
lhaendler: How about onConflict.userObjectExists = Replace | Keep?

pkg/apis/kubebind/v1alpha1/permissionclaims_types.go
Danil-Grigorev: So an object created on the provider side and synchronized on the consumer side will behave the same way as the object created on the consumer side and synchronized on the provider side is idempotent when it comes to deletion? Basically the `FollowDeletion` or `Follow` scenario, is supported by the api?

lhaendler: An object that is created by the provider and synchronized to the consumer could be donated using `ownerTransfer=Donate`. In that case, if the consumer deletes it, it will be deleted on the provider cluster too. Since donation means that the consumer will own the object, the provider may have to claim update fields to continue writing to the object after creation in the consumer cluster

PR #186, opened 2023-10-24
update README.md with the step to run the konnector

effi-ofer:
The README is missing the step to run the konnector.


README.md
```diff
@@ -110,4 +110,5 @@ WQh88mNOY0Z3tLy1/WOud7qIEEBxz+POc4j8BsYenYo=
 The `--cookie-signing-key` option is required and supports 32 and 64 byte lengths.
 The `--cookie-encryption-key` option is optional and supports byte lengths of 16, 24, 32 for AES-128, AES-192, or AES-256.
 
+* start the connector: `bin/konnector`
```
sttts, line 113: the connector is controlled by the kubectl plugin. So you don't have to run it manually. So this is included in the following line. Worth to clarify?

effi-ofer:
Thanks for the clarification.

PR #188, opened 2023-11-02
Add ClusterScopedIsolation to APIServiceExportSpec

waltforme:
This PR begins to better address the isolation of the cluster-scoped service objects. As a [follow up](https://github.com/kube-bind/kube-bind/pull/175#issuecomment-1721074910) to a previous PR.

This PR
- adds a `ClusterScopedIsolation` field to the spec of `APIServiceExport`;
- exposes this field as a flag of the example-backend;
- initializes this field when the "kube-bind-example-backend-serviceexportrequest" controller creates an APIServiceExport object.

This PR hasn't added other controllers' reactions to this field, yet.

### DIFF
--- a/contrib/example-backend/options/options.go
+++ b/contrib/example-backend/options/options.go
@@ -41,13 +41,14 @@ type Options struct {
 type ExtraOptions struct {
 	KubeConfig string
 
-	NamespacePrefix       string
-	PrettyName            string
-	ConsumerScope         string
-	ExternalAddress       string
-	ExternalCAFile        string
-	ExternalCA            []byte
-	TLSExternalServerName string
+	NamespacePrefix        string
+	PrettyName             string
+	ConsumerScope          string
+	ClusterScopedIsolation string
+	ExternalAddress        string
+	ExternalCAFile         string
+	ExternalCA             []byte
+	TLSExternalServerName  string
 
 	TestingAutoSelect string
 }
@@ -77,9 +78,10 @@ func NewOptions() *Options {
 		Serve:  NewServe(),
 
 		ExtraOptions: ExtraOptions{
-			NamespacePrefix: "cluster",
-			PrettyName:      "Example Backend",
-			ConsumerScope:   string(kubebindv1alpha1.NamespacedScope),
+			NamespacePrefix:        "cluster",
+			PrettyName:             "Example Backend",
+			ConsumerScope:          string(kubebindv1alpha1.NamespacedScope),
+			ClusterScopedIsolation: string(kubebindv1alpha1.IsolationPrefixed),
 		},
 	}
 }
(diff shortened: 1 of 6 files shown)


pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
```diff
@@ -90,8 +90,25 @@ type APIServiceExportSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="informerScope is immutable"
 	InformerScope Scope `json:"informerScope"`
+
+	// ClusterScopedIsolation specifies how cluster scoped service objects are isolated between multiple consumers on the provider side.
+	// It can be "Prefixed", "Namespaced", or "None".
+	ClusterScopedIsolation Isolation `json:"clusterScopedIsolation"`
```
sttts, line 96: is this required or can we define a default? Depends a lot on the backend. So maybe a default is not good, but rather force the backend to set this explicitly. So maybe required would be good if `apiextensionsv1.ResourceScope` below is set to `Cluster`. CEL can do this.

This also depends on whether the

pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
```diff
@@ -90,8 +90,25 @@ type APIServiceExportSpec struct {
...
 }
 
+type Isolation string
```
sttts, line 99: use kubebuilder enum marker to define which values are valid in OpenAPI.

contrib/example-backend/options/options.go
```diff
@@ -113,13 +116,22 @@ func (options *Options) Complete() (*CompletedOptions, error) {
...
 
-	// normalize the scope
+	// normalize the scope and the isolation
 	if strings.ToLower(options.ConsumerScope) == "namespaced" {
 		options.ConsumerScope = string(kubebindv1alpha1.NamespacedScope)
 	}
 	if strings.ToLower(options.ConsumerScope) == "cluster" {
 		options.ConsumerScope = string(kubebindv1alpha1.ClusterScope)
 	}
+	if strings.EqualFold(options.ClusterScopedIsolation, string(kubebindv1alpha1.IsolationPrefixed)) {
```
sttts, line 126: learned a new function: EqualFold :)

sttts:
```golang
switch options.ClusterScopedIsolation {
case "prefixed": options.ClusterScopedIsolation = kubebindv1alpha1.IsolationPrefixed
...
}
```

That's easier to read.

waltforme: Yes! Fixed this in the added commit.

pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
waltforme: Done in the latest commit.

pkg/apis/kubebind/v1alpha1/apiserviceexport_types.go
waltforme: Thanks! I realized that the `ClusterScopedIsolation` field is required _only_ when the resource is cluster-scoped, otherwise, the field should not appear in an APIServiceExport object. So I made a few changes in the added commit to improve.

Regarding defaulting of the field, I’m trying to follow. Are you referring to https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting, so that we rely on the schema to set a default value for the field? If that’s the case, we still need somewhere in code (which can be the backend, IMO) to overwrite the default when we want a non-default value, no?

waltforme: Yes, that makes sense.

I think CEL validations for the 1st line is there already. So I added that for the 2nd line in [the last commit](https://github.com/kube-bind/kube-bind/pull/188/commits/17659680e35c35f27d85127693e3e222f1a8f4cb).

If my understanding is still not aligned, please let me know, thank you!

sttts (approved):

PR #191, opened 2023-11-16
Add guide for a local kube-bind setup using kind

a9s-nc:
I've found myself jumping through some hoops to get kube-bind running all locally on my laptop the past few days.

So, hoping it will be helpful to others I've written up the steps as a small guide.

Step 3 in the "Setup" section is a bit of a pain point, I'm happy about any suggestions how to avoid specifying the host IP there.


lhaendler:
Works on my machine ™️

lhaendler (approved):

PR #195, opened 2023-11-28
Add healthcheck endpoint for konnector

lhaendler:
Add an endpoint for health checking to the konnector.

This gives the consumer information about the state of the konnector, which might help with debugging


cmd/konnector/cmd/konnector.go
```diff
@@ -33,6 +35,8 @@ import (
 	bindversion "github.com/kube-bind/kube-bind/pkg/version"
 )
 
+const LeaderElectionTimeout = 3 * time.Second
```
sttts, line 38: isn't this a pretty short period? What's the default?

sttts:
Nit about the leader election timeout. Otherwise sgtm.

sttts (approved):

cmd/konnector/cmd/konnector.go
lhaendler: k8s uses 20 seconds https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-controller-manager/app/controllermanager.go#L206

mjudeikis:
@sttts merge? :)

Issue #202, opened 2024-12-03
example-backend: generated kubeconfigs are missing CA for kube-binder service account


gman0:
When example-backend runs inside a Pod, it generates kubeconfig Secrets for kube-binder ServiceAccount without CA. 

https://github.com/kube-bind/kube-bind/blob/b3ac8f4a7db4c7eba81e23b7dc52cdba3790c74c/contrib/example-backend/kubernetes/resources/kubeconfig.go#L49-L51

InCluster kubeconfigs pass the CA by a file, not by contents which means `clusterConfig.CAData` will be empty. We need to check for `CAFile` too, and read the certificate from there.

I'll have a PR ready shortly.

PR #203, opened 2024-12-03
example-backend: when generating kubeconfigs, read CAFile if CAData is empty

gman0:
CAData is empty when example-backend is running in cluster. Instead, CA for kube-binder service account is referred to by a file. So, try to read that file if CAData is empty.

Fixes: https://github.com/kube-bind/kube-bind/issues/202


mjudeikis:
/lgtm
/approve 
i know I dont have access but still :) 
@sttts wanna take a stab to this?

sttts (approved):

PR #204, opened 2024-12-07
✨  split apis and clients to SDK sub go module

mjudeikis:
This PR is bit unexpected (surprise!), but it does the following:
1. Fixes code generation (it was broken, the same issue we had in kcp)
2. Add separate submodule with `sdk/apis` and `sdk/k8s/clients` (in the next PR will add `sdk/kcp/...` as 3rd sub-modules.

Why? When creating backend for KCP I got into very strange place where I was doing all these crazy code generations for resourceschemas, sdk's, etc. And was thinking "I just need easily importable sdk"... So here it comes.

Regards point 2 - we might not want to add it, but I dont see anything bad if it does not pollute the dependency tree of main repo?


sttts:
Would rather follow kcp's naming scheme of `sdk/client/clientset`. Adding `sdk/kcp/clientset` later IMO is fine too.

mjudeikis:
Sure, can do that. 

Just a note, tried to add kcp client and apparently our generator on kcp side does not like `kube-bind` as group with dash and we somehow not honoring `groupGoName` annotations :/ so this is getting very rabbit-holly :D

sttts (approved):

PR #219, opened 2024-12-16
Pin/revert all generators to 0.31.0

mjudeikis:
Generators by default used 0.32.0 version of k/k. This breaks both k8s and kcp sdk if used with lower version of k/k (0.31.0) due to missing ` undefined: rest.CodecFactoryForGeneratedClient` method.

This reverts generators and SDK to use 0.31.0 generators.

Plan:
1. Revert this and cut a release so we can have 0.31.0 version full support.
2. Make separate PR to bump 0.32.0 generators (maybe wait for KCP rebase to land into code-generator so we can do both at the same time)
3. Cut a release so we have both versions supported. 

If we don't do this revert now, we will not be able to support 0.31.0 k/k :/ I missed this as this missing dep was surfaced only when trying to use SDK with 1.31 kubernetes backend :/


go.mod
```diff
@@ -152,3 +154,13 @@ require (
...
+
+replace (
+	k8s.io/api => k8s.io/api v0.31.0
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.0
+	k8s.io/apimachinery => k8s.io/apimachinery v0.31.0
+	k8s.io/apiserver => k8s.io/apiserver v0.31.0
+	k8s.io/client-go => k8s.io/client-go v0.31.0
+	k8s.io/code-generator => k8s.io/code-generator v0.31.0
+	sigs.k8s.io/structured-merge-diff/v4 => sigs.k8s.io/structured-merge-diff/v4 v4.4.1
+)
```
sttts, line 166: am not a fan of these. Why do we have to override? We should not.

mjudeikis: k, gonna try force to use 0.31 deps. Tests were failing as it defaulted to 0.32 :/

sttts:
Wondering whether publishing the client-go clients and informers together with the apis is actually good idea. The APIs are pretty much independent of the kube version. The clients and informers might not.

mjudeikis:
> Wondering whether publishing the client-go clients and informers together with the apis is actually good idea. The APIs are pretty much independent of the kube version. The clients and informers might not.

but this would mean 4th go module?
```
vX.Y.Z
sdk/clients/vX.Y.Z
sdk/kcp/vX.Y.Z
sdk/apis/vX.Y.Z
```
?

go.mod
```diff
@@ -152,3 +152,6 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.17.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 )
+
+// This is needed for update-codegen-clients.sh to pin the version of code-generator.
+replace k8s.io/code-generator => k8s.io/code-generator v0.31.0
```
sttts, line 157: is this really effective?

sttts: I.e. isn't the go install in the makefile enough?

mjudeikis: We can move this, but these would need to be "refactored" a bit:
https://github.com/kube-bind/kube-bind/blob/main/hack/update-codegen-clients.sh#L25-L33

```
CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; go list -f '{{.Dir}}' -m k8s.io/code-generator)}

source "${CODEGEN_PKG}/kube_codegen.sh"
```

currently they are upstream style 🤷

sdk/apis/third_party/conditions/apis/conditions/v1alpha1/register.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// SchemeGroupVersion is group version used to register these objects
+var SchemeGroupVersion = schema.GroupVersion{Group: "conditions.kcp.io", Version: "v1alpha1"}
```
sttts, line 24: what is this for?

go.mod
mjudeikis: And this is root module, not APIS or clients. 
and once we move to 1.32 this will go away.

sdk/apis/third_party/conditions/apis/conditions/v1alpha1/register.go
mjudeikis: kcp SDK has this generated for `applyconfiguration` 
```
v1alpha1 "github.com/kube-bind/kube-bind/sdk/apis/third_party/conditions/apis/conditions/v1alpha1"

// Group=conditions, Version=v1alpha1
case v1alpha1.SchemeGroupVersion.WithKind("Condition"):
```

And we added this into kcp at some point. Just adding this here too.

mjudeikis: Ah! kcp.io group. I think its artificial, so we could make it kube-bind.io. But this is kcp SDK spefic

mjudeikis: https://github.com/kube-bind/kube-bind/blob/main/sdk/kcp/applyconfiguration/utils.go#L41

go.mod
sttts: oh, that are fake kinds? wtf

sttts (approved):

PR #225, opened 2024-12-18
Move CLI into separate module

mjudeikis:
As CLI is the definition of protocol, it should be as independent as possible.
This moves it to a separate go module so decoupling it from the main codebase.

This will help with automatic docs generation too.


sttts:
xref https://github.com/kcp-dev/kcp/pull/3084:

1. The kubectl plugins will be used in go generate setups, and there vendoring is common and the binary is called via go run. For that, having dependencies into a kube fork is strange.
2. The kubectl plugins do not need the kube forks today.
3. The kubectl plugins should only use official APIs. The PR enforces hygiene.

All these apply here too (with s/kube fork/kcp/), I agree.

sttts (approved):

mjudeikis:
> Do you want to bring the go.mod back?

Will try tomorrow :) will ping for re-approval.

sttts (approved):

sttts (approved):

PR #231, opened 2025-01-02
Add kcp backend example

mjudeikis:
This add basic kube-bind kcp backend example.

This is not the final state of this. It is still missing documentation of how to get something running (same mangodb for now, but we need a better example). 
Added a small Roadmap section too to specify what is planned next. 

ps. This is monster PR, Any ideas how to make it smaller?


mjudeikis:
superceded by https://github.com/kube-bind/kube-bind/pull/275

PR #232, opened 2025-01-04
Fix local make build targets, cli and SDK

mjudeikis:
Fix `make build` local target in the new multi-module setup.
Fix cli bind regression. 
make SDK generation non-yeat specific to avoid generation when new year comes again in few days :)


Makefile
```diff
@@ -18,6 +18,7 @@ SHELL := /usr/bin/env bash -e
 GO_INSTALL = ./hack/go-install.sh
 
 TOOLS_DIR=hack/tools
+ROOT_DIR=$(abspath .)
```
sttts, line 21: we have GOBIN_DIR below. One is enough.

cli/cmd/kubectl-bind/cmd/kubectlBind.go
```diff
@@ -70,5 +70,5 @@ func KubectlBindCommand() *cobra.Command {
 	}
 	bindCmd.AddCommand(apiserviceCmd)
 
-	return root
+	return bindCmd
```
sttts, line 73: wrong commit?

sttts:
I hate this Makefile shell magic. But I don't see a better way 🙂‍↕️

cli/cmd/kubectl-bind/cmd/kubectlBind.go
mjudeikis: not quite. I was missing one more commit to add `bindCmd` to `rootCmd`. added now

sttts (approved):

PR #243, opened 2025-02-07
📖 docs: add initial developer documentation

dakraus:
## Summary

The purpose of this PR is to add some initial developer documentation to provide insight into how the konnector works and how a kube-bind client interacts with the provider backend. At the moment, the pages mainly contain diagrams that show

* the interaction between a kube-bind client and the provider backend during cluster and API binding
* the set of controllers implemented in the konnector

I tried to describe the interaction between a kube-bind client and the provider backend in a general way - e.g. some actions like installing the konnector, which is currently implemented in the kubectl plugin, are not shown in the diagrams. Also, not all of the konnector controllers are documented yet.

Overall I would love to hear your feedback, if this is something that adds value to the kube-bind docs and could be a starting point for kube-bind backend implementors.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/content/.pages
+++ b/docs/content/.pages
@@ -3,4 +3,5 @@ nav:
     - index.md
   - Setup: setup
   - Contributing: contributing
+  - Developers: developers
   - Reference: reference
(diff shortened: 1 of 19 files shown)


mjudeikis:
I think it has enough details just to merge and there is some factual mismatches we identify later, we fix after.

mjudeikis (approved):

sttts (approved):

PR #261, opened 2025-05-09
Refactor existing APIs to use APIResourceSchema, BoundAPIResourceSchema for service exposure

cnvergence:
## Summary

- move existing apis to v1alpha2 and adapt them for new APIResourceSchema and BoundAPIResourceSchema.
- adapt konnector code
- adapt example-backend code
- fix e2e tests

## Related issue(s)

Fixes https://github.com/kube-bind/kube-bind/issues/255

## Release Notes

```release-note
Refactor existing APIs to use APIResourceSchema, BoundAPIResourceSchema for service exposure .
```


hack/kind-example-setup/app-cluster.sh
```diff
@@ -17,8 +17,12 @@
 set -o nounset
 set -o pipefail
 
-source "$(dirname "$0")/host-ip.sh"
-get_host_ip
+DEFAULT_KONNECTOR_IMAGE="ghcr.io/kube-bind/konnector:v0.4.6"
```
mjudeikis, line 20: This will break as soon as we merge this. repo has "latest" image, should we do the un-holy thing and just use latest here?

cnvergence: I wasn't aware of that, for now, I wanted a default for  "sane" tested version

mjudeikis: But test uses APISchemas already, but this implementation does not support it :D

cnvergence: No, I meant for running the example-backend kind setup with 0.4.8 version of images and kube-bind cli.
We can use the latest as the default, I don't mind

Issue #267, opened 2025-08-04
Refactor KubeBind to work with Multicluster-runtime


mjudeikis:
This is kinda follow-up for https://github.com/kube-bind/kube-bind/pull/231

We want tp refactor kube-bind to work with multicluster-runtime https://github.com/kubernetes-sigs/multicluster-runtime 
So we can do:
1. Support any backend (single cluster, or many clusters ones) as long as they work with multicluster-runtime
2. Use controller-runtime pattern

This will NOT make the konnector controller-runtime specific just to limit scope. This is an issue for that:  https://github.com/kube-bind/kube-bind/issues/280

mjudeikis:
And last PR for this is https://github.com/kube-bind/kube-bind/pull/276 and will get merged until EOW.

PR #274, opened 2025-08-08
Migrate all backend controllers to controller-runtime

mjudeikis:
## Summary

Migrate all controllers to controller-runtime

## What Type of PR Is This?
/kind cleanup
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Migrate backend to controller-runtime
```

PR #275, opened 2025-08-09
Move backend to Multicluster runtime

mjudeikis:
## Summary

Build ontop of https://github.com/kube-bind/kube-bind/pull/274
Moves backend to use multicluster-runtime. 
Move manager to use multicluster-runtime
Change API urls to be more extentable in multicluster world. 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes https://github.com/kube-bind/kube-bind/issues/267

## Release Notes

```release-note
Move to use MultiCluster Runtime
```


contrib/example-backend/config.go
```diff
@@ -58,22 +49,18 @@ func NewConfig(options *options.CompletedOptions) (*Config, error) {
...
-		return nil, err
+	switch options.Provider {
+	case "kcp":
+		provider, err := apiexport.New(config.ClientConfig, apiexport.Options{})
+		if err != nil {
+			return nil, fmt.Errorf("error setting up kcp provider: %w", err)
+		}
+		config.Provider = provider
+	default:
+		config.Provider = nil
```
cnvergence, line 62: do we consider adding up provider in the future?

contrib/example-backend/http/handler.go
```diff
@@ -277,40 +280,46 @@ func (h *handler) handleCallback(w http.ResponseWriter, r *http.Request) {
...
 
 func (h *handler) handleResources(w http.ResponseWriter, r *http.Request) {
 	logger := klog.FromContext(r.Context()).WithValues("method", r.Method, "url", r.URL.String())
 
 	prepareNoCache(w)
 
+	clusterPath := mux.Vars(r)["cluster"]
+	var cluster string
+	if clusterPath == "-" {
+		cluster = "" // empty strings means - all cluster for Multicluster-runtime. But this does not work for bind UI.
```
cnvergence: we could change it in the followup

cli/pkg/kubectl/base/kubeconfig.go
```diff
@@ -35,11 +35,11 @@ func ParseRemoteKubeconfig(kubeconfig []byte) (host string, ns string, err error
 		return "", "", nil
 	}
 	if _, found := config.Contexts[config.CurrentContext]; !found {
-		return "", "", fmt.Errorf("current context %q of remote kubeconfig not found", config.CurrentContext)
+		return "", "", fmt.Errorf("current context '%q' of remote kubeconfig not found", config.CurrentContext)
```
xrstf: This will lead to errors like

> current context '"foo"' of ...

Why quote twice?

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
```diff
@@ -87,82 +86,87 @@ func NewClusterBindingReconciler(
...
 				}
 				return binding, nil
 			},
-			getClusterRoleBinding: func(ctx context.Context, name string) (*rbacv1.ClusterRoleBinding, error) {
+			getClusterRoleBinding: func(ctx context.Context, cache cache.Cache, name string) (*rbacv1.ClusterRoleBinding, error) {
 				var binding rbacv1.ClusterRoleBinding
 				key := types.NamespacedName{Name: name}
 				if err := cache.Get(ctx, key, &binding); err != nil {
-					return nil, err
+					return nil, fmt.Errorf("failed to get ClusterRoleBinding %q: %w", name, err)
```
xrstf: Appreciate te effort, but this will lead to mostly redundant-sounding error strings.

```
ns, err := c.getClusterRoleBinding(yaddayadda)
if err != nil && !errors.IsNotFound(err) {
	return fmt.Errorf("failed to get ClusterRoleBinding %s: %w", name, err)
}
```

will now yield

> failed to get ClusterRoleBinding foo: failed to get ClusterRoleBinding "foo": clusterrolebinding.rbac.v1 "foo" not found

Since there is only 1 function call being wrapped here, I would argue the added error context in the fmt.Errorf() is unnecessary.

</nitpick>

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
```diff
@@ -199,13 +187,21 @@ func mapServiceExportToClusterBinding(ctx context.Context, obj client.Object) []
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-func (r *ClusterBindingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *ClusterBindingReconciler) Reconcile(ctx context.Context, req mcreconcile.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 	logger.Info("Reconciling ClusterBinding", "namespace", req.Namespace, "name", req.Name)
```
xrstf: Can we instead try `log.Info("Reconciling ClusterBinding", "request", req)`, so the cluster is also included?

Also, can we use a scoped logger like

```go
logger := log.FromContext(ctx).With("request", req)
logger.Info("Reconciling ClusterBinding")
```

so that all following log statements are tagged accordingly?

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
```diff
@@ -219,14 +215,14 @@ func (r *ClusterBindingReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		logger.Error(err, "Failed to reconcile ClusterBinding")
 		return ctrl.Result{}, err
 	}
 
 	// Update status if it has changed
 	if !reflect.DeepEqual(original, clusterBinding) {
-		err := r.Update(ctx, clusterBinding)
+		err := client.Update(ctx, clusterBinding)
 		if err != nil {
 			logger.Error(err, "Failed to update ClusterBinding status")
```
xrstf, line 227: This log might be redundat, IIRC controller-runtime already logs reconciling errors.

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
```diff
@@ -238,16 +234,38 @@ func (r *ClusterBindingReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		Named(controllerName).
 		Complete(r)
 }
+
+func mapCRD(clusterName string, cl cluster.Cluster) handler.TypedEventHandler[client.Object, mcreconcile.Request] {
+	return handler.TypedEnqueueRequestsFromMapFunc[client.Object, mcreconcile.Request](func(ctx context.Context, obj client.Object) []mcreconcile.Request {
+		logger := log.FromContext(ctx)
+		serviceExport, ok := obj.(*kubebindv1alpha2.APIServiceExport)
+		if !ok {
+			logger.Error(fmt.Errorf("object is not an APIServiceExport"), "unexpected object type")
```
xrstf: <nitpick>

This error should only occur if the developer made a mistake. There is nothing an administrator could ever do to fix this.

IMHO things like this should either be silently ignored or lead to a panic.

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
```diff
@@ -238,16 +234,38 @@ func (r *ClusterBindingReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return nil
+		}
+		return []mcreconcile.Request{
+			{
+				Request: reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Namespace: serviceExport.Namespace,
+						Name:      serviceExport.Name,
+					},
```
xrstf: `NamespacedName: ctrlruntimeclient.ObjectKeyFromObject(serviceExport)`

contrib/example-backend/controllers/serviceexport/serviceexport_controller.go
```diff
@@ -99,48 +96,28 @@ func NewAPIServiceExportReconciler(
...
 //+kubebuilder:rbac:groups=kubebind.k8s.io,resources=apiserviceexports/finalizers,verbs=update
 //+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-func (r *APIServiceExportReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *APIServiceExportReconciler) Reconcile(ctx context.Context, req mcreconcile.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
-	logger.Info("Reconciling APIServiceExport", "namespace", req.Namespace, "name", req.Name)
+	logger.Info("Reconciling APIServiceExport", "cluster", req.ClusterName, "namespace", req.Namespace, "name", req.Name)
```
xrstf: Same note regarding log context applies here :)

contrib/example-backend/controllers/serviceexport/serviceexport_controller.go
```diff
@@ -171,13 +148,43 @@ func (r *APIServiceExportReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	return ctrl.Result{}, nil
 }
 
+// getCRDMapper returns a mapper function that uses the manager to find related APIServiceExports.
+func getCRDMapper(clusterName string, cl cluster.Cluster) handler.TypedEventHandler[client.Object, mcreconcile.Request] {
+	return handler.TypedEnqueueRequestsFromMapFunc[client.Object, mcreconcile.Request](func(ctx context.Context, obj client.Object) []mcreconcile.Request {
+		crd := obj.(*apiextensionsv1.CustomResourceDefinition)
+		crdKey := crd.Name
+		// Get all clusters and search for APIServiceExports that might be related to this CRD
```
xrstf: I don't understand/see the `get all clusters` part here. This seems to operate on a single cluster?

contrib/example-backend/controllers/serviceexport/serviceexport_controller.go
```diff
@@ -171,13 +148,43 @@ func (r *APIServiceExportReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		var exports kubebindv1alpha2.APIServiceExportList
+		if err := c.List(ctx, &exports, client.MatchingFields{indexers.ServiceExportByCustomResourceDefinition: crdKey}); err != nil {
+			return []mcreconcile.Request{}
+		}
+
+		var requests []mcreconcile.Request
+		for _, export := range exports.Items {
+			requests = append(requests, mcreconcile.Request{
+				Request: reconcile.Request{
+					NamespacedName: types.NamespacedName{
```
xrstf: ObjectKeyFromObject(export)

contrib/example-backend/http/handler.go
xrstf: Instead of this, could we simply have two routes using the same handler func? `/clusters/{cluster}/exports` and `/exports`? This `-` feels very magiccky.

contrib/example-backend/kubernetes/resources/rbac.go
```diff
@@ -20,29 +20,176 @@ import (
...
+		if !policyRuleEqual(a[i], b[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+// policyRuleEqual compares two PolicyRule structs for equality.
+func policyRuleEqual(a, b rbacv1.PolicyRule) bool {
+	if !stringSliceEqual(a.APIGroups, b.APIGroups) {
```
xrstf: `slices.Equal()` exists :-)

contrib/example-backend/options/options.go
```diff
@@ -100,6 +103,8 @@ func (options *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&options.ExternalCAFile, "external-ca-file", options.ExternalCAFile, "The external CA file for the service provider cluster. If not specified, service account's CA is used.")
 	fs.StringVar(&options.TLSExternalServerName, "external-server-name", options.TLSExternalServerName, "The external (TLS) server name used by consumers to talk to the service provider cluster. This can be useful to select the right certificate via SNI.")
 
+	fs.StringVar(&options.Provider, "multicluster-runtime-provider", options.Provider, "The multicluster runtime provider. If not specified, the default (single cluster) is used.")
```
xrstf: Ironically "single cluster" is not a supported value for this flag. Can we have a better way of managing the possible values?

```go
var providerAliases = map[string]string{
   "kcp": "kubernetes",
   "kubernetes": "kubernetes",
   "": "kubernetes",
}

fs.StringVar(&options.Provider, "multicluster-runtime-provider", options.Provider,
	fmt.Sprintf("The multicluster runtime provider. Possible values are: %v", sets.List(sets.StringKeySet(providerAliases))),
)

resolved, ok := providerAliases[options.Provider]
if !ok { error }
```

contrib/example-backend/options/options.go
```diff
@@ -181,5 +185,12 @@ func (options *CompletedOptions) Validate() error {
 		}
 	}
 
+	switch options.Provider {
+	case "kcp", "kubernetes":
+		// all good.
+	default:
+		return fmt.Errorf("unknown multicluster-runtime provider %q", options.Provider)
```
xrstf:
```go
fmt.Errorf("unknown provider %q, must be one of %v", options.Provider, sets.List(sets.StringKeySet(providerAliases)))
```

contrib/example-backend/server.go
```diff
@@ -74,6 +71,33 @@ func NewServer(ctx context.Context, c *Config) (*Server, error) {
...
+	if err := apiextensionsv1.AddToScheme(scheme); err != nil {
+		return nil, fmt.Errorf("error adding apiextensions scheme: %w", err)
+	}
+	if err := kubebindv1alpha2.AddToScheme(scheme); err != nil {
+		return nil, fmt.Errorf("error adding kubebind scheme: %w", err)
+	}
+
+	opts := ctrl.Options{
+		Controller: config.Controller{
+			SkipNameValidation: ptr.To(true), // TODO(mjudeikis): Debug why this makes tests fail.
```
xrstf: I think in this usecase it is perfectly reasonable to disable the name check. Maybe just a `// during testing, many servers are started and stopped` comment is enough. Certainly not a TODO for the future, as there is pretty much nothing else we can do here. Except maybe add a SkipNameValidation field to the Server's Config so it can be toggled explicitly in the tests.

test/e2e/bind/happy-case_test.go
```diff
@@ -368,8 +368,8 @@ func simulateBrowser(t *testing.T, authURLCh chan string, resource string) {
 	err := browser.Open(authURL)
 	require.NoError(t, err)
 
-	t.Logf("Waiting for browser to be at /resources")
-	framework.BrowerEventuallyAtPath(t, browser, "/resources")
+	t.Logf("Waiting for browser to be at /clusters/-/resources")
+	framework.BrowerEventuallyAtPath(t, browser, "/clusters/-/resources")
```
xrstf: Wanna fix the typo in `BrowerEventuallyAtPath` and make it `BrowserEventuallyAtPath`? :)

xrstf (changes requested):
Mostly style stuff :)

contrib/example-backend/http/handler.go
mjudeikis: we could. This is basically google API standard. google doing this. But yap, good point.

cli/pkg/kubectl/base/kubeconfig.go
mjudeikis: easier to read when its something like "default". you know when something is "of interest" in the error.

xrstf: But why wrap the string in single quotes AND double quotes? Why not `""""%q""""` to highlight it even more? Adding multiple quotes will make me squint harder to see why there are so many '''''''''''' all of a sudden.

Also, this is completely inconsistent and most other code still just uses %q -- which is perfectly fine, since %q quotes already.

contrib/example-backend/http/handler.go
xrstf: "Google API Standard"? Never heard of that, and that after decades of being a webdev.

I find the "-" to be extremely confusing and completely unnecessary hacky magic to solve something that can be done much cleaer and more logical by just having 2 routes.

mjudeikis: https://cloud.google.com/apis/design and https://google.aip.dev/1 
Basically, we've been using this in few gigs I did when you need to "agree wider organization API standard". Take this, fork it, adjust and you have API reference. And I grew custom to it as its easy, well documented and easy to reuse. And rest and GRPC compatible. 

It just be my social bubble thingy :) 

```
message ListConfigsRequest {
  // To list all configs, use `-` as the user id.
  // Formats:
  // * `users/-`
  // * `users/{user}`
  //
  // Note: Specifying an actual user id will return a collection of one config.
  // Use GetConfig instead.
  string parent = 1 [
    (google.api.resource_reference).child_type = "api.googleapis.com/Config"];

  // other standard pagination fields...
}
```

This is one of the examples &&

contrib/example-backend/config.go
mjudeikis: Maybe, but nil is - single cluster context.

contrib/example-backend/http/handler.go
mjudeikis: In the end this is why we have reviews :) Will move towards many paths, same handler :)

contrib/example-backend/controllers/clusterbinding/clusterbinding_controller.go
mjudeikis: > Appreciate te effort

Are you really? :D I accept straightforward comments :) You can turn off this filter with me if you want to 😉

contrib/example-backend/controllers/serviceexport/serviceexport_controller.go
mjudeikis: This leaked from other PR sources further down the chain. removing

contrib/example-backend/kubernetes/resources/rbac.go
mjudeikis: AI driven refactoring :D

contrib/example-backend/server.go
mjudeikis: make sense. good one.

deploy/konnector/90-deployment.yaml
```diff
@@ -32,4 +32,4 @@ spec:
         readinessProbe:
           httpGet:
             path: "/healthz"
-            port: 8080
+            port: 8090
```
mjudeikis, line 35: This changed due to local `quickstart` not working as backend and konnector uses same port. And its just for probles, so its ok.

PR #285, opened 2025-08-19
Slim down cookies

### LINKED ISSUE #136: `securecookie: the value is too long` error when using more complex encryption algorithms for oAuth
I configured `kube-bind` against [KeyCloak](https://www.keycloak.org/), and when doing oAuth I get the following error:

```I1108 10:58:30.731079   59508 handler.go:282] "failed to encode secure session cookie" method="GET" url="/callback?state=xxx&session_state=xxxx" error="securecookie: the value is too long"```

It appears that depending on the encryption algorithm used for oAuth, `AccessToken` and `RefreshToken` can grow quite long, pushing the overall cookie size beyond the allowed 4096 bytes. https://github.com/kube-bind/kube-bind/blob/48f693686ff5aafdcd5a032e8e3f977e37d20b58/contrib/example-backend/http/handler.go#L265-L274

I got around the problem by removing `RefreshToken` from the list of stored elements in the cookie. The oAuth algorithm can also be altered to generate a smaller token size, but that may not be an option for all scenarios. I wonder if all the info stored in the cookie is needed, or can we for example remove the `RefreshToken` all together?

understandably it is a PoC so this is not anything urgent. just highlighting it here for others and for future improvements.

### PR DESCRIPTION
**Pull request by xrstf**
## Summary
Depending on the crypto settings, but more likely depending on the number of groups a user is in, the size of the generated access, refresh and ID tokens can be quite large. However browser cookies are limited to 4KB per origin.

Thankfully kube-bind's example backend doesn't actually need any of the large tokens, neither does it need the user groups. So for this implementation it's possible to simply store much less in the cookie.

This is what this PR achieves. We now only store the issuer and subject from the original ID token. When testing using the quickstart guide, the size of my cookie went down from 2,476 to 492 bytes. And I suspect it won't grow much beyond that, unless your issuer is https://examplesareoftennotenoughsometimesyouneedawizardtoexplainthebasicsofmagictoyou.edu.

Since the cookie is now much smaller and hopefully stays small, I have removed the example compression functions, as well as the example (?) function to parse the cookie SameSite attribute. Not sure why that one ever existed.

I also made the name of the cookie less random, but still per-cluster, so that fewer information needs to be transmitted via query string parameters. There is IMHO no need to have multiple sessions open for the same cluster and the same backend.

For the cookie settings, I have removed the TODO for the HttpOnly flag. With the example backend there is absolutely no reason to make this actually configurable. The same applies to the SameSite flag, which is purely dependent on how the example backend is implemented and doesn't need to be configurable. Likewise I extracted the secure flag, in preparation of supporting HTTPS mode for the example backend if we ever need it.

I also fixed a few log statements that were using Info instead of Error.

And lastly, noticed during testing, I fixed `make dex/bin/dex`.

## What Type of PR Is This?
/kind bug

## Related Issue(s)
Fixes #136

## Release Notes
```release-note
The example backend stores less data in the session cookie and should be more compatible with a variety of OIDC provider setups.
```


contrib/example-backend/session/cookie.go
```diff
@@ -14,39 +14,22 @@ See the License for the specific language governing permissions and
...
 	"net/http"
 	"time"
 )
 
-func MakeCookie(req *http.Request, name string, value string, expiration time.Duration) *http.Cookie {
+func MakeCookie(req *http.Request, name string, value string, secure bool, expiration time.Duration) *http.Cookie {
 	return &http.Cookie{
 		Name:     name,
 		Value:    value,
 		Path:     "/", // TODO: make configurable
```
cnvergence: we might also drop all the TODOs

cnvergence (approved):

contrib/example-backend/session/cookie.go
xrstf: I was thinking that someone might expose this backend behind an Ingress with a PathPrefix, so the TODO for the Path felt valid.

PR #294, opened 2025-09-03
Feat/helper function in dynamic schema parsing

### LINKED ISSUE #293: Add helper function in dynamic schema parsing
Currently, schemas are parsed in not a very type-safe way:

Example:
```
scope := item.UnstructuredContent()["spec"].(map[string]interface{})["scope"]
```

We should add a helper function to do this better

RyanCarlisle:
/kind feature

Fixes #293

## Release Notes

```release-note
NONE
```

### DIFF
--- a/backend/http/handler.go
+++ b/backend/http/handler.go
@@ -378,31 +378,57 @@ func (h *handler) handleResources(w http.ResponseWriter, r *http.Request) {
 
 	var result []UISchema
 	for _, item := range apiResourceSchemas.Items {
-		scope := item.UnstructuredContent()["spec"].(map[string]interface{})["scope"]
+		spec, err := ConvertInterfaceToMapStringInterface(item.UnstructuredContent()["spec"])
+		if err != nil || spec == nil {
+			logger.Error(err, "failed to get data from unstructured content")
+			http.Error(w, "internal error", http.StatusInternalServerError)
+			return
+		}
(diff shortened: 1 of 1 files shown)


backend/http/handler.go
```diff
@@ -378,31 +378,57 @@ func (h *handler) handleResources(w http.ResponseWriter, r *http.Request) {
 
 	var result []UISchema
 	for _, item := range apiResourceSchemas.Items {
-		scope := item.UnstructuredContent()["spec"].(map[string]interface{})["scope"]
+		spec, err := ConvertInterfaceToMapStringInterface(item.UnstructuredContent()["spec"])
```
mjudeikis, line 381: I would imagine this being something like https://github.com/crossplane/crossplane-runtime/blob/main/pkg/fieldpath/fieldpath.go 

Where you basicaly say:
`fieldpath.Get(unstructured, "spec.group.names") and it will return them safely in a single call.

RyanCarlisle:
Hi @mjudeikis , could you elaborate more on fieldpath.Get()? I checked the code in the link you provided, but I didn't see a definition of the Get() in fieldpath.

mjudeikis:
`field, ok, err := unstructured.NestedFieldCopy(item.UnstructuredContent(), "spec", "group", "names")` is better option I think.

PR #295, opened 2025-09-04
Move to schemaless backend

mjudeikis:
## Summary

In this PR we do the following:
1. Make backend full schemaless - no need hard schema to serve crds. As long as object is compliant with crd schema (like kcp apiresourceschemas) - its ok.
2. Removes kube-bind APIResourceSchema as its not needed anymore
3. Rename BoundAPiResourceSchema to `BoundSchema`. purspose is same - track status of bound schema
4. Make APIServiceBinding 1:1 to APIServiceExport - this was an oversight in previous refactors. 

## What Type of PR Is This?
/kind api-change
/kind cleanup
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Refactor backend to be fully schemaless and clean api as a result
```

PR #304, opened 2025-09-11
Add PermissionClaims

mjudeikis:
## Summary

This adds the ability to claim related resources. For now, it's ServiceAccounts, ConfigMaps and Secrets.
Claims are label-based.
The origin of the resources determines the owner. 

High-level design:

1. Permission Claims are specified using `GVR` type syntax, same as resource Claims.

2. LabelSelectors are used to scope down the resources. Based on label selectors, informers will be constructed.
Label selector is immutable and should be the provider lead. This means the provider determines what it is, and the consumer has no say in the matter. In the future, we could allow consumers to say, but we would need to ship admissionWebhook. And I want to avoid this complexity now.

3. Permission on provider side - Its either `*` on the namespace level (when isolation is namespaced, or same on cluster level). 

This is mainly as verbs dont quite make sense in kube-bind sense. As you always need more than what might imply. In the end, Kube verbs are ineffective because they are based on informers, requiring specific permissions (list, get, watch) to be used. Since the permission model only applies to `konnector` to `provider` only, even if the provider states it needs `read` (list, watch, get), it is not applicable. 

I suggest deferring permission to future enhancement if needed, and even then, make them super flat (read, write, all). But for now, let's keep the API super simple. 

4.  Ownership is defined by the fact of where the object was created first - consumer/provider. In the future, we can define "take over conditions". For example, when the Provider seeds the configuration, but the consumer takes over. But this is even possible now by just updating the label to swap owners.  Label is automatically added by the reconciler if not specified `kube-bind.io/owner=provider/consumer`

To see this in action, follow https://github.com/kube-bind/kube-bind/blob/main/contrib/kcp/README.md from the PR.

Permission claim example:
```yaml
apiVersion: kube-bind.io/v1alpha2
kind: APIServiceExportRequest
metadata:
  name: cowboys.wildwest.dev
spec:
  resources:
  - group: wildwest.dev
    resource: cowboys
    versions:
    - v1alpha1
  permissionClaims:
  - apiGroup: ""
    resource: configmaps
    selector:
      labelSelector:
        matchLabels:
          app: wildwest
  - apiGroup: ""
    resource: secrets
    selector:
      namedResource:
      - name: wildwest-secrets
        namespace: bob
```

## What Type of PR Is This?
/kind feature
/kind api-change

## Related Issue(s)

~~Based on https://github.com/kube-bind/kube-bind/pull/295 and it must merge first~~
Fixes: https://github.com/kube-bind/kube-bind/issues/256

## Release Notes

```release-note
Add basic permission claims to API
```

## Summary by CodeRabbit

* **New Features**
  * Permission-claims for selective resource syncing (label & named-resource selectors); exposed across exports, requests, bindings, CRDs, and status.
  * Per-export RBAC provisioning, multiple new controllers (claiming/syncing, namespace provisioning), centralized controller lifecycle store, claimable-API resolver, and UI now consumes typed exported schemas.
  * New helper for determining claimed resources and test client helper.

* **Bug Fixes**
  * Legacy namespace identity annotations auto-migrated; RBAC rule added for bound-schema status updates.

* **Documentation**
  * Updated contrib/kcp guides and examples.

* **Tests**
  * Expanded e2e coverage and new unit tests for selectors and namespace handling.

* **Chores**
  * CRD schema/version/storage updates; improved E2E parallelism control and tooling exclusions.

### DIFF
--- a/sdk/apis/kubebind/v1alpha2/apiserviceexportrequest_types.go
+++ b/sdk/apis/kubebind/v1alpha2/apiserviceexportrequest_types.go
@@ -104,6 +104,9 @@ type APIServiceExportRequestSpec struct {
 	// +required
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:XListType=map
+	// +kubebuilder:validation:XListMapKey=resource
+	// +kubebuilder:validation:XListMapKey=group
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="resources are immutable"
 	Resources []APIServiceExportRequestResource `json:"resources"`
 


Makefile
```diff
@@ -114,7 +114,7 @@ all: build
 check: verify lint test test-e2e
 .PHONY: check
 
-GOMODS := $(shell find . -name 'go.mod' -exec dirname {} \; | grep -v hack/tools)
```
ntnn, line 121: Do we need dex here anyhow? The Makefile currently has two ways to get dex - once via tools in hacks/tools/dex-$ver and once in dex/
Do we need both?

backend/controllers/servicenamespace/servicenamespace_reconcile.go
```diff
@@ -75,6 +78,131 @@ func (c *reconciler) reconcile(ctx context.Context, client client.Client, cache
...
+		permissions := []rbacv1.PolicyRule{}
+		for _, claim := range export.Spec.PermissionClaims {
+			permissions = append(permissions, rbacv1.PolicyRule{
+				APIGroups: []string{claim.Group},
+				Resources: []string{claim.Resource},
+				// We need list and watch for informers to be able to start. And create to create initial object.
+				Verbs: []string{"*"},
+			})
+		}
+		if c.scope == kubebindv1alpha2.ClusterScope {
```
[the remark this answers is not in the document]
mjudeikis: We need to discuss this one. I think this simplifies things and APIs in a way. 
This permission only valid on provider Namespace level. namespace is already owned by consumer only so there is no issues of leaking things.
And without this you run into issues like if you give only Read permisisons, consumer still need to have Create to create initial secret if one comes from consumers.

I just dont think having this granular in kube-bind scope make sense.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -155,44 +140,48 @@ func (r *reconciler) ensureControllers(ctx context.Context, name string, export
...
-			schemaName := parts[0]
-			if !processedSchemas[schemaName] {
-				logger.V(1).Info("Stopping APIServiceExport resource sync", "key", key, "reason", "Schema no longer referenced")
-				c.cancel()
-				delete(r.syncContext, key)
-			}
+	// Ensure controller for permission claims
+	if err := r.ensureControllersForPermissionClaims(ctx, export, binding, isClusterScoped); err != nil {
+		errs = append(errs, err)
+	}
```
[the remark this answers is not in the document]
mjudeikis: All claims should be same scoped. We dont allow mixed scopes.

Makefile
mjudeikis: In the one of readme, we say gitclone, build, and use.
once we close it its there and local runs just create noise.

We could unify how and were we install it. But as followup

pkg/konnector/controllers/cluster/serviceexport/spec/spec_reconcile.go
```diff
@@ -62,6 +63,9 @@ func (r *reconciler) reconcile(ctx context.Context, obj *unstructured.Unstructur
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      ns,
 					Namespace: r.providerNamespace,
+					OwnerReferences: []metav1.OwnerReference{
+						*metav1.NewControllerRef(r.apiServiceExport, kubebindv1alpha2.SchemeGroupVersion.WithKind("APIServiceExport")),
+					},
 				},
```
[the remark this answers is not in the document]
mjudeikis: nil checkign is above in the code

sdk/apis/kubebind/v1alpha2/boundchema_types.go
```diff
@@ -41,6 +46,13 @@ type BoundSchema struct {
 	Status BoundSchemaStatus `json:"status,omitempty"`
 }
 
+// ResourceGroupName returns the group name of the resource.
+//
+// Important: If you change this, change one for APIServiceExportRequestResource too.
+func (b *BoundSchema) ResourceGroupName() string {
+	return fmt.Sprintf("%s.%s", b.Spec.Names.Plural, b.Spec.Group)
+}
```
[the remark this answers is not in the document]
mjudeikis: This is used only for CRDs so groups should never be empty

pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
```diff
@@ -0,0 +1,447 @@
...
+				// TODO: This is not quite right place for this code...
+				_, err := c.createServiceNamespace(context.TODO(), &kubebindv1alpha2.APIServiceNamespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      ns,
+						Namespace: c.providerNamespace,
+						OwnerReferences: []metav1.OwnerReference{
+							*metav1.NewControllerRef(c.apiServiceExport, kubebindv1alpha2.SchemeGroupVersion.WithKind("APIServiceExport")),
+						},
+					},
+				})
```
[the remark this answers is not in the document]
mjudeikis: This controller does not run if .apiServiceExport is not set

pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
```diff
@@ -0,0 +1,485 @@
...
+				if err != nil && !errors.IsAlreadyExists(err) {
+					runtime.HandleError(fmt.Errorf("failed to create APIServiceNamespace %q: %w", ns, err))
+					return
+				}
+				// Requeue when the APIServiceNamespace is created.
+				logger.V(2).Info("created APIServiceNamespace, requeueing", "namespace", ns)
+				// Once APIServiceNamespace is created, the informer will pick it up and requeue the objects.
+				return
+			}
```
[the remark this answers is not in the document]
mjudeikis: This one is a hack, I need to find a way to fix it in a better way before this is merged. But reviews can be done already

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
```diff
@@ -72,97 +79,70 @@ func (r *reconciler) reconcile(ctx context.Context, cl client.Client, cache cach
...
-			if !ok || err != nil || group == "" {
-				klog.FromContext(ctx).Error(err, "Skipping invalid schema: missing group", "ns", item.GetNamespace(), "name", item.GetName())
-				schemaFailed = true
-			}
-			plural, ok, err := unstructured.NestedString(obj, "spec", "names", "plural")
-			if !ok || err != nil || plural == "" {
-				klog.FromContext(ctx).Error(err, "Skipping invalid schema: missing names.plural", "ns", item.GetNamespace(), "name", item.GetName())
-				schemaFailed = true
-			}
+	var boundSchemas kubebindv1alpha2.ExportedSchemas = make(map[string]*kubebindv1alpha2.BoundSchema, len(list.Items))
```
ntnn:
```suggestion
	boundSchemas := make(kubebindv1alpha2.ExportedSchemas, len(list.Items))
```
Just a style nit. I was thinking `ExportedSchemas` was an interface for a second.

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
```diff
@@ -72,97 +79,70 @@ func (r *reconciler) reconcile(ctx context.Context, cl client.Client, cache cach
...
-						req.Name, group, plural, scope, r.informerScope,
-					)
-					req.Status.Phase = kubebindv1alpha2.APIServiceExportRequestPhaseFailed
-					req.Status.TerminalMessage = conditions.GetMessage(req, kubebindv1alpha2.APIServiceExportRequestConditionExportsReady)
-					// We can't proceed with this request.
-					return fmt.Errorf("resource %s/%s has scope %q which is incompatible with backend informer scope %q", group, plural, scope, r.informerScope)
-				}
+	// Ensure all bound schemas exist
+	for _, res := range req.Spec.Resources {
+		if len(res.Versions) == 0 {
```
ntnn, line 140: Shouldn't this compare with exportedSchemas still?
E.g. if all versions were cleaned out but there are versions in the existing exported schemas?

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
```diff
@@ -283,3 +264,85 @@ func (r *reconciler) ensureExports(ctx context.Context, cl client.Client, cache
...
+					kubebindv1alpha2.APIServiceExportRequestConditionExportsReady,
+					"DifferentScopes",
+					conditionsapi.ConditionSeverityError,
+					"Different scopes found: %v",
+					scopes,
+				)
+				return fmt.Errorf("different scopes found for claimed resources: %v", scopes)
+			}
+		}
+	}
```
ntnn:
```suggestion
    first := apiextensionsv1.ResourceScope("")
	for _, res := range req.Spec.Resources {
		boundSchema, ok := exportedSchemas[res.ResourceGroupName()]
		if !ok {
			conditions.MarkFalse(
				req,
				kubebindv1alpha2.APIServiceExportRequestConditionExportsReady,
				"SchemaNotFound",
				conditionsapi.ConditionSeverityError,
				"Schema %s not found",
				res.ResourceGroupName(),
			)
			return fmt.Errorf("schema %s not found", res.ResourceGroupName())
		}
        if first == apiextensionv1.ResourceScope("") {
            first = boundSchema.Spec.Scope
            continue
        } 
        if boundSchema.Spec.Scope != first {
			conditions.MarkFalse(req,
				kubebindv1alpha2.APIServiceExportRequestConditionExportsReady,
				"DifferentScopes",
				conditionsapi.ConditionSeverityError,				
                "Different scopes found: %v",
				scopes,
			)
            return fmt.Errorf("different scopes found for claimed resources: %v", scopes)
        }
	}
```

sdk/apis/kubebind/v1alpha2/apiserviceexportrequest_types.go
```diff
@@ -106,6 +106,12 @@ type APIServiceExportRequestSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="resources are immutable"
 	Resources []APIServiceExportRequestResource `json:"resources"`
+
+	// PermissionClaims records decisions about permission claims requested by the service provider.
+	// Access is granted per GroupResource.
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Permission claim selector is immutable"
```
ntnn:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="permissionClaims are immutable"
```

Since the list should be immutable.

Also doesn't this need a maplist and key to enforce immutability?

pkg/konnector/controllers/contextstore/contextstore.go
```diff
@@ -0,0 +1,131 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package contextstore
+
+// contextstore allows to manage and track context per controllers, stored at the different levels of the controller hierarchy:
+// APIServiceExport - schemas and permissionClaims - each schema runs its own gvr controller and each permissionClaim runs its own controller.
+// Context are stored at the APIServiceExport level.
```
ntnn:
```suggestion
// contextstore allows to manage and track context per controllers, stored at the different levels of the controller hierarchy:
// APIServiceExport - schemas and permissionClaims - each schema runs its own gvr controller and each permissionClaim runs its own controller.
// Context are stored at the APIServiceExport level.
package contextstore
```

So go recognizes it as a package docstring. I don't think the comment on its own is recognized as package doc.

pkg/konnector/controllers/contextstore/contextstore.go
```diff
@@ -0,0 +1,131 @@
...
+	var cancel func()
+	c.lock.Lock()
+	if ctx, ok := c.store[key]; ok {
+		cancel = ctx.Cancel
+		delete(c.store, key)
+	}
+	c.lock.Unlock()
+	if cancel != nil {
+		cancel()
+	}
```
ntnn:
```suggestion
	c.lock.Lock()
	if ctx, ok := c.store[key]; ok {
		defer ctx.Cancel()
		delete(c.store, key)
	}
	c.lock.Unlock()
```

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -288,25 +279,203 @@ func (r *reconciler) ensureControllerForSchema(ctx context.Context, export *kube
...
+	if err != nil {
+		cancel()
+		return err
 	}
 
+	logger.Info("creating claim reconciler", "gvr", claimGVR, "key", claimKey)
+
+	// Start the informers and controllers in a goroutine
+	go func() {
+		defaultConsumerInf.Start(ctxWithCancel.Done())
```
ntnn, line 476:
```suggestion
	go func() {
        defer cancel()
		defaultConsumerInf.Start(ctxWithCancel.Done())
```

So the context gets cleaned up even if the controller exits early.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -288,25 +279,203 @@ func (r *reconciler) ensureControllerForSchema(ctx context.Context, export *kube
...
+	binding *kubebindv1alpha2.APIServiceBinding,
+	claim kubebindv1alpha2.PermissionClaim,
+	export *kubebindv1alpha2.APIServiceExport,
+	claimGVR runtimeschema.GroupVersionResource,
+	isClusterScoped bool,
+	claimKey contextstore.Key,
+) error {
+	logger := klog.FromContext(ctx)
+
+	ctxWithCancel, cancel := context.WithCancel(ctx)
```
ntnn: I think `ctxWithCancel` could be inititaed just before starting the goroutine.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
ntnn: I think it would also make sense to defer a `r.syncStore.Delete` here instead to ensure that the "outer" controller can restart the controller when it does crash because the entry is gone from the syncstore.

pkg/konnector/controllers/cluster/serviceexport/spec/spec_reconcile.go
```diff
@@ -50,6 +51,9 @@ type reconciler struct {
 // reconcile syncs downstream objects (metadata and spec) with upstream objects.
 func (r *reconciler) reconcile(ctx context.Context, obj *unstructured.Unstructured) error {
 	logger := klog.FromContext(ctx)
+	if r.apiServiceExport == nil { // Should never happen, but we check to make sure we dont regress in the future.
+		panic("apiServiceExport is nil")
```
ntnn: I know this shouldn't happen but I really don't like panics :D
Can we return an error?

pkg/konnector/controllers/cluster/serviceexport/multinsinformer/informer.go
```diff
@@ -184,7 +188,21 @@ func (inf *DynamicMultiNamespaceInformer) enqueueServiceNamespace(obj any) {
...
+			if err != nil {
+				utilruntime.HandleError(fmt.Errorf("failed to convert label selector: %w", err))
+				return
+			}
+			options.LabelSelector = selector.String()
+		}
+		factory = dynamicinformer.NewFilteredDynamicSharedInformerFactory(inf.providerDynamicClient, time.Minute*30, sns.Status.Namespace, tweakListOptions)
+	} else {
+		factory = dynamicinformer.NewFilteredDynamicSharedInformerFactory(inf.providerDynamicClient, time.Minute*30, sns.Status.Namespace, nil)
+	}
```
ntnn:
```suggestion
    var tweakListOptions func(options *metav1.ListOption) = nil
	if inf.labelSelector != nil {
		tweakListOptions := func(options *metav1.ListOptions) {
			selector, err := metav1.LabelSelectorAsSelector(inf.labelSelector)
			if err != nil {
				utilruntime.HandleError(fmt.Errorf("failed to convert label selector: %w", err))
				return
			}
			options.LabelSelector = selector.String()
		}
    }

	factory = dynamicinformer.NewFilteredDynamicSharedInformerFactory(inf.providerDynamicClient, time.Minute*30, sns.Status.Namespace, tweakListOptions)
```

pkg/konnector/controllers/cluster/claimedresources/claimedresources_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	updateProviderObject func(ctx context.Context, obj *unstructured.Unstructured) error
+	deleteProviderObject func(ctx context.Context, ns, name string) error
+
+	getConsumerObject    func(ctx context.Context, ns, name string) (*unstructured.Unstructured, error)
+	updateConsumerObject func(ctx context.Context, obj *unstructured.Unstructured) (*unstructured.Unstructured, error)
+	createConsumerObject func(ctx context.Context, ob *unstructured.Unstructured) (*unstructured.Unstructured, error)
+	deleteConsumerObject func(ctx context.Context, ns, name string) error
+}
+
+// reconcile syncs conmsumer claimed resources to provider.
```
ntnn:
```suggestion
// reconcile syncs consumer claimed resources to provider.
```

pkg/konnector/controllers/cluster/claimedresources/claimedresources_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+			r.makeProviderOwner(candidate)
+
+			if _, err := r.createConsumerObject(ctx, candidate); err != nil {
+				return err
+			}
+
+			return nil
+		}
+
+		if providerObj.GetDeletionTimestamp() != nil && !providerObj.GetDeletionTimestamp().IsZero() {
```
ntnn:
```suggestion
		if !providerObj.GetDeletionTimestamp().IsZero() {
```

`.IsZero` checks for nilness

pkg/konnector/controllers/cluster/claimedresources/claimedresources_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	candidate.SetLabels(labels)
+
+	annotations := map[string]string{}
+	for key, annotation := range obj.GetAnnotations() {
+		if strings.Contains(key, "kcp.io/cluster") {
+			continue
+		}
+		annotations[key] = annotation
+	}
+	candidate.SetAnnotations(annotations)
```
ntnn, line 231: Maybe we should have a list of keys to wipe somewhere and e.g. just wipe any labels and annotations from `kcp.io`, `kubernetes.io`?
E.g. kubectl apply sets a label (I think, might also be an annotation) to track changes. Similar for argocd et al. I don't think we want these to be transferred.

Or maybe it would be better to have an allowlist controlled by the user?

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
mjudeikis: Request are short lived and but yes. We have all full story to implement lifecycle of boundschemas.

mjudeikis: It just out of scope for this one for now.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
mjudeikis: r.syncStore.Delete will do cancel too, so once is enough. I think tis is what you had in mind

pkg/konnector/controllers/cluster/claimedresources/claimedresources_reconciler.go
mjudeikis: I think this is a follow-up. We have a ticket for metadata. In general, we will need more of these:
1. Add routing details
2. Remove some of these ^^ as you suggest. 

Didin't wanted to blow this out of controll :/

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -87,22 +85,17 @@ func (r *reconciler) reconcile(ctx context.Context, name string, export *kubebin
...
-			if strings.HasSuffix(key, "."+name) {
-				logger.V(1).Info("Stopping APIServiceExport sync", "key", key, "reason", "APIServiceExport deleted")
-				c.cancel()
-				delete(r.syncContext, key)
-			}
+		deleted := r.syncStore.BulkDeletePrefixed(exportKey)
+		for _, k := range deleted {
+			logger.V(1).Info("Stopping APIServiceExport sync", "key", k.Key(), "reason", "NoAPIServiceExport")
 		}
+
```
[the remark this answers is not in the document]
mjudeikis: bulk delete already cancels inside.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -155,44 +141,48 @@ func (r *reconciler) ensureControllers(ctx context.Context, name string, export
...
-			r.lock.Unlock()
+		if c.Generation == export.Generation {
 			return nil // all as expected
 		}
 
 		logger.V(1).Info("Stopping APIServiceExport resource sync", "key", key, "reason", "GenerationChanged", "generation", schema.Generation)
-		c.cancel()
-		delete(r.syncContext, key)
+		r.syncStore.Delete(key)
 	}
```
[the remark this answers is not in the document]
mjudeikis: Delete already cancels.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -288,25 +280,226 @@ func (r *reconciler) ensureControllerForSchema(ctx context.Context, export *kube
...
+
+	// Cleanup controllers for claims that are no longer present
+	claimPrefix := contextstore.NewKey(export.Namespace, export.Name, "claim")
+	contexts := r.syncStore.ListPrefixed(claimPrefix)
+	for _, c := range contexts {
+		if !processedClaims[c.Key().String()] {
+			logger.V(1).Info("Stopping permission claim controller", "key", c.Key(), "reason", "Claim no longer present")
+			r.syncStore.Delete(c.Key())
+		}
 	}
```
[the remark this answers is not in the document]
mjudeikis: delete already cancels

pkg/konnector/controllers/cluster/serviceexport/serviceexport_controller.go
```diff
@@ -75,14 +76,19 @@ func NewController(
...
+
 	c := &controller{
 		queue: queue,
 
 		serviceExportLister:  serviceExportInformer.Lister(),
 		serviceExportIndexer: serviceExportInformer.Informer().GetIndexer(),
 
-		serviceNamespaceLister:  serviceNamespaceInformer.Lister(),
-		serviceNamespaceIndexer: serviceNamespaceInformer.Informer().GetIndexer(),
+		//	serviceNamespaceLister:  serviceNamespaceInformer.Lister(),
```
cnvergence: are we dropping this lister and indexer or just temp?

mjudeikis: was not used. removing

pkg/konnector/controllers/cluster/claimedresourcesnamespaces/claimedresourcesnamespaces_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	providerNamespace string
+
+	// Channel to notify claimedresources controller when new namespaces are created
+	// so we don't have to wait for the next resync period.
+	claimedResourcesEnqueueChan chan<- string
+}
+
+func (c *controller) isClaimed(obj *unstructured.Unstructured) bool {
+	return resources.IsClaimed(c.claim.Selector, obj)
+}
```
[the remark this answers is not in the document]
mjudeikis: Will remove `*` for now as its un-validated complexity

backend/controllers/servicenamespace/servicenamespace_reconcile.go
```diff
@@ -75,6 +76,156 @@ func (c *reconciler) reconcile(ctx context.Context, client client.Client, cache
...
+					rb := rolebinding.DeepCopy()
+					rb.Subjects = expectedSubjects
+					rb.RoleRef = rbacv1.RoleRef{Kind: "Role", Name: name, APIGroup: "rbac.authorization.k8s.io"}
+					if err := client.Update(ctx, rb); err != nil {
+						return fmt.Errorf("failed to update RoleBinding %s: %w", name, err)
+					}
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mjudeikis: Added follow-up issue for this https://github.com/kube-bind/kube-bind/issues/329

sdk/apis/kubebind/v1alpha2/apiserviceexportrequest_types.go
```diff
@@ -106,6 +106,12 @@ type APIServiceExportRequestSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="resources are immutable"
 	Resources []APIServiceExportRequestResource `json:"resources"`
+
+	// PermissionClaims records decisions about permission claims requested by the service provider.
+	// Access is granted per GroupResource.
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="permissionClaims are immutable"
+	PermissionClaims []PermissionClaim `json:"permissionClaims,omitempty"`
```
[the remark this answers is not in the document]
ntnn: I think I mentioned that in comment earlier as well. To make a list immutable the backing type should be a map, otherwise if e.g. element one and two are exchanged the list would be seen as not equal for validation even though only the order of elements changed.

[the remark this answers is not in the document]
mjudeikis: I think I seen it it just got lost in noise.

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
```diff
@@ -283,3 +264,106 @@ func (r *reconciler) ensureExports(ctx context.Context, cl client.Client, cache
...
+			req,
+			kubebindv1alpha2.APIServiceExportRequestConditionExportsReady,
+			"SchemaNotFound",
+			conditionsapi.ConditionSeverityError,
+			"Schema not found",
+		)
+		return fmt.Errorf("no exported schemas found")
+	}
+
+	first := apiextensionsv1.ResourceScope("")
```
cnvergence, line 291: nit: we could just use `var first apiextensionsv1.ResourceScope`

pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
```diff
@@ -0,0 +1,525 @@
...
+	bindlisters "github.com/kube-bind/kube-bind/sdk/client/listers/kubebind/v1alpha2"
+)
+
+const (
+	controllerName = "kube-bind-konnector-claimed-object"
+)
+
+// NewController returns a new controller reconciling downstream objects to upstream.
+func NewController(
+	//	scope kubebindv1alpha2.InformerScope,
```
cnvergence, line 54: let's drop the scope

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -288,25 +280,226 @@ func (r *reconciler) ensureControllerForSchema(ctx context.Context, export *kube
...
+			claim.Selector.LabelSelector,
+		)
+
+		if err != nil {
+			logger.Info("aborting", "error", err)
+			return err
+		}
+	}
+
+	serviceNamespaceChannel := make(chan string, 100)
```
cnvergence, line 438: do we need to close it?

mjudeikis: It should be alive as long as process is alive. If process is dead - nothing to close :) In general - yes, but its just noop type of closing

backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
mjudeikis: will address in follow-up

pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
mjudeikis: will address in follow-up

PR #311, opened 2025-09-18
Various nits, fixes and cleanup

ntnn:
## Summary

1. Ensure make recipes work across all modules
2. Fix `sdk/client` importing `sdk/apis` by merging into one package `sdk`
This causes errors when the APIs change but are not present in main, because the generated client tries to import from `sdk/apis` which upstream isn't updated yet ofc.
3. Fix various fix golangci-lint found (which were hidden because lint only ran on the root module)
4. Fix code generator by updating kcp-dev/code-generator (not sure why but v2.3.1 wouldn't build due to a dependency erroring; v2.4.0 works though)
5. Tidy all modules
6. Also fix the `make help` output

This was actually intended to be a smaller PR with just the recipe changes but you know about the 99 bugs

## What Type of PR Is This?

/kind bug
/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Merge `sdk/client` and `sdk/apis` into `sdk` package
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,9 @@ LDFLAGS := \
 all: build
 .PHONY: all
 
+check: verify lint test test-e2e
+.PHONY: check
+
 GOMODS := $(shell find . -name 'go.mod' -exec dirname {} \; | grep -v hack/tools)
 
 ldflags:


mjudeikis (dismissed):
Normally I would say - let's not split `sdk/{client,api}` 
But as we are working towards converting all code to be multicluster-runtime, and hence dont need generated client anymore - `sdk/client` should go away very soon. 
So this makes code simplier for now, and easier to consume.

mjudeikis (approved):
lets ship this.

Issue #296, opened 2025-09-05
kcp e2e tests


mjudeikis:
We now gonna support more agnostic backend (like kcp) after https://github.com/kube-bind/kube-bind/pull/295 merges.

We need to add dedicated tests for it into `kcp/test/e2e` to test `kcp/README.md` described flow to make sure we dont add regressions in our schema'less backend

mjudeikis:
This was done as two days ago

PR #314, opened 2025-09-19
Move kcp/ to contrib/kcp/

ntnn:
## Summary

Just the move so actualy changes aren't lost in the noise later.

I was considering adding a `test-contribs` recipe, but the unit test recipe is already module aware so that would be overkill.
The `test-e2e` should stay as it is as it needs the setup with kcp and dex, which may not be required for all contribs tests.

There's also the `build` recipe that directly targets `kcp-init`. I've left it for now, I'm not yet sure if the contribs/ should be self-contained or if the makefile should build absolutely everything.
Same for codegen - I could see `codegen` looking into the makefiles of the contribs and execute a codegen target if it exists. At the moment the codegen script calls into the kcp makefile explicitly.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


Makefile
```diff
@@ -314,7 +325,7 @@ modules: ## Run go mod tidy to ensure modules are up to date
 .PHONY: verify-modules
 verify-modules: modules  # Verify go modules are up to date
 	@for MOD in $(GOMODS); do \
-		(cd $$MOD; echo "Verifying $$MOD"; if ! git diff --quiet HEAD -- go.mod go.sum; then echo "[$$MOD] go modules are out of date, please run 'make modules'"; exit 1; fi; ) \
+		(cd $$MOD; echo "Verifying $$MOD"; if ! git diff --quiet HEAD -- go.mod go.sum; then git diff -- go.mod go.sum; echo "[$$MOD] go modules are out of date, please run 'make modules'"; exit 1; fi; ) \
```
ntnn, line 328: Added this because a verify CI job failed without any indication what is different. This way we'll at least see why the modules are different.

cnvergence (dismissed):

cnvergence:
Is the e2e test scenario a part of https://github.com/kube-bind/kube-bind/issues/296?

PR #355, opened 2025-11-03
Simplify Makefile and tool installations

xrstf:
## Summary
This introduces [µget](https://codeberg.org/xrstf/uget) to download build tools and replace the `go-install.sh`. µget is better because it can download prebuilt binaries and handles checksums to detect malicious upstream changes. For a developer, nothing really changes in the workflows ("make lint" works just as before), only when you update a tool will you have to run `UGET_UPDATE=true make .......` to make µget update all checksums.

Besides this, this PR also performs some general cleanup to our Makefiles:

* no more seemingly unused `mktemp -d` for each `make` invocation
* `GOBIN_DIR` was renamed to `BUILD_DIR`, because `TOOLS_GOBIN_DIR` was confusing. Now it's `TOOLS_DIR` and `BUILD_DIR`, plus their `ABS_...` variants.
* The `INSTALL_GOBIN` logic was removed, since it was also unused.
* Build tools are no longer installed into `bin/`, but into `hack/tools/`, clearly separating them from our own build artifacts.
* The `tools.go` was thinned out a lot. Since we already, even before this PR, installed most these tools using a fake module with versions sourced from our Makefile, having synthetic Go dependencies in kube-bind makes no sense (it would only make sense if somewhere we did an unversioned `go install k8s.code-generator/cmd/lister-gen`) except for the kube codegen where we refer to its kube_codegen.sh by `go list`ing the module.
* `make clean` was added to remove `BUILD_DIR` and `TOOLS_DIR`.

## What Type of PR Is This?
/kind cleanup

## Release Notes
```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Centralized build and tool outputs into dedicated build/tool directories, added clean target, and adjusted build/install flows.
  * Introduced a guarded downloader with checksum verification and manifest, removed the legacy Go installer, and switched tool acquisition to the new downloader or clone-and-build flows; removed several now-unused module dependencies.

* **Documentation**
  * Bumped Go requirement to 1.24+, renamed quickstart title to "kube-bind", and updated setup docs to use the new make-based tool build/start workflow and clarified local setup steps.

### DIFF
--- a/docs/content/setup/.pages
+++ b/docs/content/setup/.pages
@@ -2,5 +2,6 @@ nav:
     - index.md
     - quickstart.md
     - local-setup-with-kind.md
+    - kcp-setup.md
     - helm.md
     - kubectl-plugin.md


docs/content/setup/kcp-setup.md
```diff
@@ -53,7 +52,7 @@ staticClients:
 Start dex:
 
 ```bash
-./bin/dex serve examples/config-dev.yaml
+./hack/tools/dex serve examples/config-dev.yaml
```
ntnn, line 55:
```suggestion
make run-dex
```

I don't think examples/ exists, or?
Anyhow the `run-dex` recipe could be used instead.

xrstf: Good point.

`run-dex` however uses our own internal development config, whereas the docs mean to instruct the user to use dex's config files. However since the `make dex` command now builds the dex binary and then deletes the temporary clone, that `config-dev.yaml` (https://github.com/dexidp/dex/blob/master/examples/config-dev.yaml) is not available anymore.

What would you suggest? Revert the docs to make people continue to manually clone and build _and not delete_ dex, or copy dex's `config-dev.yaml` into our repo? Or copy it too when building dex (i.e. make our Makefile copy the file)?

ntnn: Mhm.. I think we can handwave the dex configuration and just use the development config.
When people implement kube-bind in their environment they will have to configure and use their OIDC provider anyhow, explicitly requiring the user to configure dex doesn't really add anything IMHO.

Issue #349, opened 2025-10-24
Add embedded oidc


mjudeikis:
Currently, it's hard to "just get running" as OIDC is always required.

kcp has embedded etcd https://github.com/kcp-dev/embeddedetcd.
We should implement something similar for kube-bind with OIDC. 

Suggested flow to explore:
1. Have a lightweight Dex wrapper in Go as a separate command target (potentially in a separate Go module to avoid dependency pollution).
2. If OIDC is not provided, add it as a separate container in the pod and configure kube-bind to use it with some default built-in provider.
3. Persistence - either use Kubernetes or a volume with Sqlite.

PR #368, opened 2025-11-07
Embedded OIDC for quickstart

mjudeikis:
## Summary

This PR adds an internal mock OIDC so one can use the backend without an external Dex. Tests are still using dex so we cover that testing corner
Adds `kubebind dev create` command to try to create "a development instance of kube-bind using two kind clusters" (preview). Will need follow-up to update released helm charts and images 

`kubectl bind dev create/delete` will require more iteration. Currently its very quick draft so we can test with as many people as we can do fix all the edge cases (and there is a lot as every workstation is different) so dont nit pick too much :D 

 

## What Type of PR Is This?
/kind feature 

## Related Issue(s)

Fixes https://github.com/kube-bind/kube-bind/issues/349 

## Release Notes

```release-note
Add embedded OIDC  --oidc-type embedded (empty is same as before - external)
Add preview command `kubectl bind dev` to create and delete local dev environment of kube-bind
```


backend/oidc/oidc.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func New(caBundleFile string, listener net.Listener, addrOverride string) (*Server, error) {
+	// Add offline_access to supported scopes for refresh token support
+	ensureOfflineAccessScope()
+	var tlsConfig *tls.Config
+	s := &Server{}
+	if caBundleFile != "" {
+		var err error
+		tlsConfig, err = LoadTLSConfig(caBundleFile)
```
ntnn:
```suggestion
		tlsConfig, err := LoadTLSConfig(caBundleFile)
```

backend/oidc/oidc.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *Server) TLSConfig() *tls.Config {
+	return s.tlsConfig
+}
+
+func (s *Server) AddRoutes(mux *mux.Router) {
+	s.server.AddRoutes(mux)
+}
+
+// URL returns the base URL of the embedded OIDC server.
+func (s *Server) Config(callbackURL, issuerURL string) (*Config, error) {
```
ntnn: This docstring is odd?

backend/oidc/oidc.go
```diff
@@ -0,0 +1,135 @@
...
+// ensureOfflineAccessScope adds "offline_access" to mockoidc's supported scopes if not already present
+func ensureOfflineAccessScope() {
+	offlineAccess := "offline_access"
+
+	// Check if offline_access is already in the supported scopes
+	for _, scope := range mockoidc.ScopesSupported {
+		if scope == offlineAccess {
+			return // Already present
+		}
+	}
```
ntnn, line 130:
```suggestion
    if slices.Contains(mockoidc.ScopesSupported, offlineAccess) {
        return // Already present
    }
```

cli/pkg/kubectl/base/browser.go
```diff
@@ -45,5 +45,10 @@ func isWindows() bool {
 
 // isMacOS checks if the current OS is macOS
 func isMacOS() bool {
-	return strings.Contains(strings.ToLower(exec.Command("uname").String()), "darwin")
+	cmd := exec.Command("uname")
+	output, err := cmd.Output()
+	if err != nil {
+		return false
+	}
+	return strings.TrimSpace(string(output)) == "Darwin"
```
ntnn, line 53: Is there a reason this is changed back?^^ RN it is equivalent to the windows one

cli/pkg/kubectl/base/kubeconfig.go
```diff
@@ -137,3 +138,12 @@ func EnsureKubeconfigSecret(ctx context.Context, kubeconfig, name string, client
 
 	return secret, false, nil
 }
+
+// LoadRestConfigFromFile loads a kubeconfig string and returns a rest.Config
+func LoadRestConfigFromFile(kubeconfig string) (*rest.Config, error) {
```
ntnn: Shoulds this be named `Load...FromFile`? If I see that correct in the use the `kubeconfig` input is supposed to be the content of a kubeconfig file, not the path to a file.

I think something like `ParseKubeconfig` or something would be more fitting.

cli/pkg/kubectl/base/browser.go
mjudeikis: Funny thing is that short version didn't work :D And I was - compiler will optimize this either way :D

Issue #365, opened 2025-11-06
fix --dry-run in bind apiservice


mjudeikis:
Currently if one does --dry-run it is just semi dry run. It will print the `APIServiceExportRequest`, but it will see the consumer cluster with the kubeconfig and namespace. This is done so that one can immediately use the returned request.

This is not a dry run as such, because we basically do more than we say. We should fix this, but for this to work, we will need to store kubeconfig and other related objects locally too, so the user can re-use them in follow-up.

Dry run should split the process into 2:
1. Establish the contract but do not do anything on consumer side. This should work even without local cluster available. 
2. Allows to pick assets from dry-run output and continuing the process (even on different cluster)

mjudeikis:
lets put this on the ice for now

PR #369, opened 2025-11-11
fix: prevent dry-run from creating resources on consumer cluster

olamilekan000:
## Summary
```
Change fixes dry-run flag to ensure that binding assets are only saved locally without accessing 
the consumer cluster, and adds `--from-dry-run` flag to apply saved dry-run sessions later.
```
image

```
making use of the assets
```
image

```
running apiservice dry-run for template
```

image

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes [365](https://github.com/kube-bind/kube-bind/issues/365)

## Release Notes

```release-note
Added dry-run mode that saves assets locally without consumer cluster access, and `--from-dry-run` flag to apply saved dry-run sessions.
```

### DIFF
--- a/cli/pkg/kubectl/base/dryrun.go
+++ b/cli/pkg/kubectl/base/dryrun.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2025 The Kube Bind Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package base
+
+import "time"
+
+// DryRunAssets represents the saved assets from a dry-run
+type DryRunAssets struct {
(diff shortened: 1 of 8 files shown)


mjudeikis:
We've been discussing with @xrstf  that this might need a bit more thinking. 
Maybe we can explisitly ~~force~~ ask the user to provide `--dry-output-file` to output this, else - instead of this `sessionID` we give base64 state into screen? Else `.kube-bind` will get polluted very fast if somebody runs this more often. 

 Open to other suggestions on how we can change this?

mjudeikis:
I think if file provided - dont base64 it, if not - output to the screen in base64? which could be used as input as it. 
Would need to model this in cli first maybe before writing more code.

olamilekan000:
> I think if file provided - dont base64 it, if not - output to the screen in base64? which could be used as input as it. Would need to model this in cli first maybe before writing more code.

okay. How soon will you be able to do the modelling?

mjudeikis:
So been thinking about this. I think simple:
```
KUBECONFIG=kind-consumer.kubeconfig kubectl bind --dry-run      
DryRun output:
<base64 encoded string>

to use the output:
```
KUBECONFIG=kind-consumer.kubeconfig kubectl bind --from-dry-run <base64 string>
```

option 2:
```
KUBECONFIG=kind-consumer.kubeconfig kubectl bind --dry-run  --dry-run-output output.yaml # non base64 at this point
and 
KUBECONFIG=kind-consumer.kubeconfig kubectl bind --from-dry-run  output.yaml
```

wdyt? do you see any issues with this way?

PR #376, opened 2025-11-11
Add refrences API

mjudeikis:
## Summary

* Add PermissionClaims.References api with jsonPath
* Make permission claims OR'ed (it was AND'ed now) 
* Fix provider to consumer object namespace syncing
* Rework kcp user-facing docs
* Refactor tests more to reuse existing examples so we have less yaml to maintain

References api example:

```yaml
  permissionClaims:
  - group: ""
    resource: secrets
    selector:
      labelSelector:
        matchLabels:
          app: sheriff
      references:
        - resource: sheriffs
          group: wildwest.dev
          jsonPath:
            name: spec.secretRefs.#.name
            namespace: spec.secretRefs.#.namespace
```

## What Type of PR Is This?
/kind feature
/kind api-change

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Reference api for PermissionClaims
Make permissionClaims OR'ed vs AND'ed now
Fix namespace re-map from provider to consumer
```

### DIFF
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -40,6 +40,114 @@ type ServiceNamespaceLister interface {
 	List(selector labels.Selector) ([]*kubebindv1alpha2.APIServiceNamespace, error)
 }
 
+func matchesLabelSelector(logger klog.Logger, selector *metav1.LabelSelector, obj *unstructured.Unstructured) bool {
+	if selector == nil {
+		return false
+	}
+
+	labelSelector, err := metav1.LabelSelectorAsSelector(selector)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("invalid label selector in permission claim: %w", err))
+		return false
+	}
+
+	objLabels := obj.GetLabels()
+	if objLabels == nil {
+		objLabels = make(map[string]string)
+	}
+
+	matches := labelSelector.Matches(labels.Set(objLabels))
+	if matches {
+		logger.Info("resource matched", "name", obj.GetName(), "namespace", obj.GetNamespace(), "claim-type", "labelSelector", "selector", labelSelector.String(), "objectLabels", objLabels)
+	}
+	return matches
+}
+
+func matchesNamedResources(logger klog.Logger, namedResources []kubebindv1alpha2.NamedResource, obj *unstructured.Unstructured) bool {
+	if len(namedResources) == 0 {
+		return false
+	}
+
+	for _, nr := range namedResources {
+		if nr.Namespace != "" && nr.Namespace != obj.GetNamespace() {
+			continue
+		}
+		if nr.Name == obj.GetName() {
+			logger.Info("resource matched", "name", obj.GetName(), "namespace", obj.GetNamespace(), "claim-type", "namedResource", "namedResource", nr)
+			return true
+		}
+	}
+	return false
+}
+
+func matchesReferences(logger klog.Logger, references []kubebindv1alpha2.SelectorReference, obj *unstructured.Unstructured, potentiallyReferencedResources *unstructured.UnstructuredList) bool {
+	if len(references) == 0 {
+		return false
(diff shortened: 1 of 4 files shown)


test/e2e/framework/backend.go
```diff
@@ -52,21 +50,7 @@ func InstallKubebindCRDs(t testing.TB, clientConfig *rest.Config) {
...
-		"--oidc-issuer-url=http://127.0.0.1:5556/dex",
-		"--cookie-signing-key=" + base64.StdEncoding.EncodeToString(signingKey),
-		"--cookie-encryption-key=" + base64.StdEncoding.EncodeToString(encryptionKey),
-	}, args...)...)
-}
-
-func StartBackendWithoutDefaultArgs(t testing.TB, args ...string) (net.Addr, *backend.Server) {
-	ctx, cancel := context.WithCancel(context.Background())
-	t.Cleanup(cancel)
+	ctx := t.Context()
```
ntnn: There's no cancelling anymore, so this line can be deleted and the `ctx` references replaced with `t.Context`

pkg/resources/resources.go
```diff
@@ -60,10 +83,177 @@ func IsClaimed(selector kubebindv1alpha2.Selector, obj *unstructured.Unstructure
...
+						}
+
+						if referenceMatches {
+							break
+						}
+					}
+				}
+			}
+		}
+	}
```
ntnn: Wouldn't it be better to place at least parts of this in a function? It seems a bit deep. And some things would read better as e.g.

```go
if !nameResults.Exists() {
    continue
}

// the rest
```

Instead of having the code that should be executed another level deeper

test/e2e/framework/backend.go
mjudeikis: This explains why I have backers lurking in the background :D

pkg/resources/resources.go
```diff
@@ -60,10 +83,177 @@ func IsClaimed(selector kubebindv1alpha2.Selector, obj *unstructured.Unstructure
...
+				// Find versions from apiServiceExport resources
+				for v := range apiServiceExport.Spec.Resources {
+					if ref.Group == apiServiceExport.Spec.Resources[v].Group && ref.Resource == apiServiceExport.Spec.Resources[v].Resource {
+						versions = apiServiceExport.Spec.Resources[v].Versions
+						break
+					}
+				}
+			}
+
+			ctx, cancel := context.WithTimeout(context.TODO(), time.Second*5)
```
cnvergence, line 207: nit: can we pass the context?
and maybe keep it with defer out of this loop

pkg/resources/resources.go
```diff
@@ -60,10 +83,177 @@ func IsClaimed(selector kubebindv1alpha2.Selector, obj *unstructured.Unstructure
...
+// IsClaimedWithReference returns true if the given object is claimed by the given permission claim.
+// It handles both consumer side (no namespace remapping) and provider side (namespace remapping via APIServiceNamespace).
+// The function fetches potentially referenced resources if a reference selector is present.
+func IsClaimedWithReference(
+	logger klog.Logger,
+	obj *unstructured.Unstructured,
+	consumerSide bool,
+	claim kubebindv1alpha2.PermissionClaim,
+	apiServiceExport *kubebindv1alpha2.APIServiceExport,
+	providerNamespace string,
```
cnvergence: is providerNamespace used here?

pkg/resources/resources.go
mjudeikis: There is no context in here :/ 
This needs to be replaced with informer, but it's challenging. this is big of the temp  hack and once dust settles we will need to come back to this. 
added TODO

pkg/resources/resources.go
mjudeikis: good cache

Issue #329, opened 2025-09-29
Add rbac cleanup logic


mjudeikis:
If export goes away, we never clean anything related to it. We should clean namespaces, clusterscoped rbac, etc.

PR #378, opened 2025-11-12
fix cleaning up of rbac resource

olamilekan000:
## Summary
```
change fixes the cleanup of associated RBAC resources when an APIServiceExport
resource gets deleted
```

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes [329](https://github.com/kube-bind/kube-bind/issues/329)

## Release Notes

```release-note
Fix the cleaning up of RBAC resources when an APIServiceExport is deleted.
```

### DIFF
--- a/backend/controllers/servicenamespace/servicenamespace_reconcile.go
+++ b/backend/controllers/servicenamespace/servicenamespace_reconcile.go
@@ -20,12 +20,17 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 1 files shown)


backend/controllers/servicenamespace/servicenamespace_reconcile.go
```diff
@@ -314,3 +322,74 @@ func (c *reconciler) listAPIServiceExports(ctx context.Context, cache cache.Cach
...
 }
+
+func (c *reconciler) cleanupRBACResources(
+	ctx context.Context,
+	client client.Client,
+	cache cache.Cache,
+	sns *kubebindv1alpha2.APIServiceNamespace,
+	apiServiceExports *kubebindv1alpha2.APIServiceExportList,
+) error {
+	expected := make(map[string]bool)
```
xrstf: I'd suggest using a `sets.Set[string]` instead of a plain map here.

backend/controllers/servicenamespace/servicenamespace_reconcile.go
```diff
@@ -314,3 +322,74 @@ func (c *reconciler) listAPIServiceExports(ctx context.Context, cache cache.Cach
...
+	for _, export := range apiServiceExports.Items {
+		name := fmt.Sprintf("kube-binder-export-%s-%s", sns.Name, export.Name)
+		expected[name] = true
+	}
+	prefix := fmt.Sprintf("kube-binder-export-%s-", sns.Name)
```
xrstf: This variable can be moved up and then also be used inside the loop to construct the object names. Reducing code duplication.

test-cleanup-export.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: kube-bind.io/v1alpha2
```
xrstf: Did you mean to commit this file?

olamilekan000: not at all. I thought i removed it.

PR #381, opened 2025-11-14
Start "usage" docs 

mjudeikis:
## Summary

Add initial usage user facing docs 

## What Type of PR Is This?
/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Start usage docs 
```

### DIFF
--- a/docs/content/usage/api-concepts.md
+++ b/docs/content/usage/api-concepts.md
@@ -1,0 +1,513 @@
+---
+title: API Concepts
+description: |
+  Deep dive into kube-bind's core API types, their relationships, and how they work together to enable cross-cluster service binding.
+weight: 210
+---
+
+# kube-bind API Concepts
+
+This guide provides a comprehensive overview of kube-bind's core API types and how they work together to enable secure, cross-cluster service binding.
+
+## Overview
+
+kube-bind uses several specialized Kubernetes custom resources to orchestrate the export and import of APIs between clusters. Understanding these types and their relationships is essential for effective use of kube-bind.
+
+## APIServiceExportTemplate
+
+**Purpose**: Defines a reusable template for exporting a group of related APIs and their permission requirements.
+
+**Used by**: Service providers  
+**Scope**: Cluster-scoped  
+**Lifecycle**: Long-lived template definition
+
+### Key Features
+
+- **Resource Grouping**: Combines multiple related CRDs into a single bindable service
+- **Permission Claims**: Specifies additional resources the service needs access to
+- **Namespace Management**: Defines namespaces that should be created on the provider & consumer side
+- **Reusability**: Can be bound by multiple consumers
+
+### Structure
+
+```yaml
+apiVersion: kube-bind.io/v1alpha2
+kind: APIServiceExportTemplate
+metadata:
+  name: my-service-template
+spec:
+  scope: Cluster # or Namespace
+  description: "A comprehensive service template"
+  
+  # Core APIs being exported
+  resources:
+  - group: example.com
+    resource: widgets
+    versions: ["v1", "v1alpha1"]
+  - group: example.com  
+    resource: gadgets
+    versions: ["v1"]
+    
+  # Additional resources the service needs access to
+  permissionClaims:
+  - group: ""
+    resource: secrets
+    selector:
+      labelSelector:
+        matchLabels:
+          component: my-service
+      references:
+      - resource: widgets
+        group: example.com
+        jsonPath:
+          name: spec.secretRef.name
+          namespace: spec.secretRef.namespace
+  
+  # Pre-created namespaces for this service. These will be owned by provider (service owned).
+  namespaces:
+  - name: my-service-system
+    description: "System namespace for service components"
+```
+
+### Template vs Instance
+
+- **Template**: The definition (APIServiceExportTemplate) - shared and reusable
+- **Instance**: The actual export (APIServiceExport) - created per binding
+
+## APIServiceExport 
+
+**Purpose**: Represents an active, instantiated export of a specific set of CRD's and permission claims to consumer clusters.
+
+**Used by**: Automatically created by konnector agents  
+**Scope**: Namespaced  
+**Lifecycle**: Created when consumers bind to templates
+
+### Key Features
+
+- **Instance Management**: One APIServiceExport per contract per binding
+- **Status Tracking**: Reports synchronization status and conditions
+- **Automatic Lifecycle**: Created/updated/deleted by the konnector agent
+
+### Structure
+
+```yaml
+apiVersion: kube-bind.io/v1alpha2
+kind: APIServiceExport
+metadata:
+  name: widgets-export-consumer123
+  namespace: kube-bind-provider-ns
+spec:
+  # The CRD being exported
+  resources:
+  - group: example.com
+    resource: widgets
+    versions: ["v1"]
+    
+  # Permission claims for this specific export
+  permissionClaims:
+  - group: ""
+    resource: secrets
+    selector:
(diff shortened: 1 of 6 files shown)


deploy/examples/cr-cowboy.yaml
```diff
@@ -1,23 +1,40 @@
...
+metadata:
+  name: wild-west
+---
 apiVersion: wildwest.dev/v1alpha1
 kind: Cowboy
 metadata:
   name: billy-the-kid
+  namespace: wild-west
 spec:
   intent: "Ride the range and protect the cattle"
```
xrstf, line 13: This is incorrect. Billy The Kid never protected cattle in his life.

docs/content/usage/api-concepts.md
```diff
@@ -0,0 +1,495 @@
...
+        matchLabels:
+          component: my-service
+      references:
+      - resource: widgets
+        group: example.com
+        jsonPath:
+          name: spec.secretRef.name
+          namespace: spec.secretRef.namespace
+  
+  # Pre-created namespaces for this service
```
xrstf: Where are these namespaces created? Provider or consumer side?

docs/content/usage/api-concepts.md
```diff
@@ -0,0 +1,495 @@
...
+    resource: secrets
+    selector:
+      labelSelector:
+        matchLabels:
+          component: widget-service
+          consumer: consumer123
+          
+  # How isolation is done at the provider side
+  clusterScopedIsolation: Prefixed
+  # Informer scope for this export
```
xrstf: I get that we're not going to explain everything in detail here, but this comment is currently adding nothing. As my latin teacher used to say

> "A Go player plays Go."
> What does he do?

docs/content/usage/api-concepts.md
```diff
@@ -0,0 +1,495 @@
...
+spec:
+  # Reference to the template being bound
+  templateRef:
+    name: widget-service-template
+    
+  # Consumer-specific configuration  
+  resources:
+  - group: example.com
+    resource: widgets
+    versions: ["v1"]
```
xrstf, line 179: What is this exactly? Why do I need to list the resources here again? Is it like "accepting" a permission claim? (in kcp speak, like only binding half the resources of an APIExport)?

docs/content/usage/api-concepts.md
```diff
@@ -0,0 +1,495 @@
...
+
+**Purpose**: Represents the consumer-side binding to a provider service, containing the applied CRDs and managing the resource synchronization.
+
+**Used by**: Automatically created by consumer konnector agents  
+**Scope**: Namespaced (on consumer side)  
+**Lifecycle**: Created when APIServiceExportRequest is processed, long-lived
+
+### Key Features
+
+- **CRD Management**: Contains the actual CRDs applied to the consumer cluster
```
xrstf: This contradicts the description for BoundSchema, which also claims to "contain the actual CRDs to be bound". I was also surprised that we're stuffing all CRDs into a single Kube object, wouldn't that easily lead to size issues?

docs/content/usage/api-concepts.md
mjudeikis: both. But if its part of the contract - provider owns them, else any namespace is consumer owned

docs/content/usage/api-concepts.md
mjudeikis: I think this is wrong now. This was from the old doc

docs/content/usage/api-concepts.md
mjudeikis: Not sure what this question means. Its basically same payload as the template, just Kind and few feelds dropped. Jut that that `templateRef:` is trash and should not exist (old data). So might be where the confusion came from

docs/content/usage/api-concepts.md
```diff
@@ -0,0 +1,495 @@
...
+```yaml  
+APIServiceBinding → consumer applies CRDs and establishes sync
+```
+
+### 5. Bidirectional Sync
+```yaml
+Provider Resources ↔ Consumer Resources (via konnector agents)
+```
+
+## APIServiceNamespace
```
xrstf: I'd move this section up by one, so all the api resources are listed under each other before we get to "more text". I know that the flow described in "Complete Binding Flow" doesn't make use of APIServiceNamespace, but still. My 3 kroner.

docs/content/usage/index.md
```diff
@@ -0,0 +1,107 @@
...
+
+### [Template References](template-references.md)
+Advanced guide for using dynamic resource selection through references in templates.
+
+## Common Workflows
+
+### For Service Providers
+1. **Create templates** defining what APIs and resources to export
+2. **Configure permission claims** specifying what additional resources are needed
+3. **Deploy konnector agents** to handle consumer connections
```
xrstf: I don't understand this at all. I thought the provider installs the _backend_, and the consumer, armed with the kube-bind kubectl plugin, runs a command to connect, which then also installs the konnector agent in the _consumer_ cluster?

docs/content/usage/index.md
```diff
@@ -0,0 +1,107 @@
...
+4. **Manage namespace isolation** and resource access
+
+### For Service Consumers  
+1. **Discover available templates** through the web UI or CLI
```
xrstf, line 63: Wouldn't step 1 be to bind to a provider first, before the list of APIs of that provider becomes available to browse?

docs/content/usage/index.md
```diff
@@ -0,0 +1,107 @@
...
+
+```
+┌─────────────────┐         ┌─────────────────┐
+│ Service Provider│         │Service Consumer │
+│                 │         │                 │
+│ ┌─────────────┐ │         │ ┌─────────────┐ │
+│ │ Templates   │ │         │ │ Bound APIs  │ │
+│ │ Exports     │ │◄────────┤ │ Requests    │ │
+│ │ Namespaces  │ │ Secure  │ │ Namespaces  │ │
+│ └─────────────┘ │Connection │ └─────────────┘ │
```
xrstf: This is misaligned.

docs/content/usage/usage
xrstf, line 1: What is this file for?

pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
xrstf, line 1: All these Go changes.... not mentioned in the PR description. Did you accidentally merge two branches before pushing?

mjudeikis: This was built ontop of previous branch, which got merged. rebased and its gone

docs/content/usage/index.md
mjudeikis: No. You can see templates first with just UI/CLI without creating binding. But missing part was authentication (so you could see them)

docs/content/usage/index.md
mjudeikis: removed

docs/content/usage/api-concepts.md
mjudeikis: updated

docs/content/usage/api-concepts.md
mjudeikis: Updated

docs/content/usage/index.md
mjudeikis: Yes, This was wrong. Bit of AI haluciantion I think :D

docs/content/usage/api-concepts.md
mjudeikis: moved

PR #394, opened 2025-11-20
Add namespaced isolation mode for cluster-scoped objects

### LINKED ISSUE #307: Implement Isolation
Currently backend has this in the API:

```
// Isolation is an enum defining the different ways to isolate cluster scoped objects
//
// +kubebuilder:validation:Enum=Prefixed;Namespaced;None
type Isolation string

const (
	// Prepends the name of the cluster namespace to an object's name.
	IsolationPrefixed Isolation = "Prefixed"

	// Maps a consumer side object into a namespaced object inside the corresponding cluster namespace.
	IsolationNamespaced Isolation = "Namespaced"

	// Used for the case of a dedicated provider where isolation is not necessary.
	IsolationNone Isolation = "None"
)
```

But in reality, `IsolationPrefixed` is only an implementation. 
We should implement Namespaced and None.

xrstf:
## Summary
This PR implements two new scoping methods for cluster-scoped objects:

* `none` doesn't change anything about the name of a consumer-side object when it's synced to the provider. This comes with all the obvious pitfalls.
* `namespaced` transforms a cluster-scoped object into a namespaced one, putting it into the cluster namespace (`kube-bind-abc123`).

However to accomplish this, I had to refactor the entire approach to how object scoping is handled in the spec/status controllers.

Originally, kube-bind was written for only namespaced objects and everything was easy. Then we added basic support for cluster-scoped objects by prepending them with a prefix. This change however was yanked into the controllers with a crowbar. For example, the function to create the copy of an object on the provider cluster (`createProviderObject`) suddenly was transforming the given object into something else, then applies it, then un-transforms it again before returning it. From what I can see in the codebase, there is never even a reason to undo any transformations being done. Once `createProviderObject` is called, we're done processing the object. Why have all the code for reverting changes? Likewise, injecting and changing namespaces was also done in the small functional wrappers (like `createProviderObject`, `updateProviderObject`) which were never meant to contain business logic, but exist (as far as I can tell) to ease unit testing and to make the reconcilers "more functional" (i.e. instead of giving them loads of small details, just give them the few pieces of behaviour they rely on). `createProviderObject` should IMHO not have anything to do with how the objects are changed during transit.

Now that new isolation modes have been added, the old approach would not have scaled anyway and so I introduced the concept of "isolation strategies". At first I had three (prefixed (the old), namespaced, none), but then when I looked through the spec/status controllers event handlers, a lot of them also contained magic to undo/unmap cluster-scoped objects. So that logic was also moved into the isolation strategies.

However this was ugly with code basically being

```
if namespaced {
   lots of code
   yadda yadda
   APIServiceNamespace handling
   requeueing...
} else {
   isolationStrategy.DoStuff(obj)
}
```

Since the entire `APIServiceNamespace` handling is exclusive to how _namespaced_ objects are handled, I was trying to avoid having it remain in the controllers and so I decided that a fourth isolation strategy makes the most sense:

* `ServiceNamespaced` is the strategy for natively namespaced (on both sides) objects.
* `Prefixed` is the old cluster-scoped behaviour.
* `Namespaced` is one of the new cluster-scoped behaviours.
* `None` is one of the new cluster-scoped behaviours.

In the CRD however we only expose the latter three, since we always use the `ServiceNamespaced` strategy for namespaced objects [at the moment...].

The only thing that remained as part of the controller logic was applying the ClusterNamespaceAnnotation, since it applies to every object no matter what strategy and can easily be done transparently to the isolation strategies. And since this annotation is always present, I awarded it its own dedicated `types` package in the konnector to make it independent of the `controllers/` packages.

**Warning:** Due to the nature of this feature, the namespaced isolation requires a **namespaced CRD on the provider side.** (and a cluster-scoped CRD on the consumer side). Since the backend currently only has 1 global CLI flag to control isolation and can otherwise just look at the existing CRDs on the provider cluster, the backend cannot distinguish between namespaced CRDs that are meant to stay namespaced, and those that are meant to be translated into cluster-scoped CRDs.

This means that if you run your backend with `--cluster-scoped-isolation=namespaced`, **ALL NAMESPACED CRDs ON THE PROVIDER CLUSTER WILL BECOME CLUSTER-SCOPED CRDs**. Currently you would need to run two separate backends if you want to offer Services that make use of this translation, and those where you want to have natively namespaced CRDs on both sides (consumer and provider).

This limitation only affects this new isolation mode, since it's the only one that has differnt scoping on different clusters.

## What Type of PR Is This?
/kind feature

## Related Issue(s)
Fixes #307

## Release Notes
```release-note
Implement support for `Namespaced` and `None` cluster isolation strategies when dealing with cluster-scoped objects.
```

### DIFF
--- a/docs/content/usage/synchronization.md
+++ b/docs/content/usage/synchronization.md
@@ -118,7 +118,7 @@ Cluster-scoped objects require a different approach to isolation than namespaced
 
 #### None Strategy
 
-The `none` strategy providers no consumer-separation at all. Any cluster-scoped object on the consumer side is copied 1:1 to the provider side.
+The `none` strategy provides no consumer-separation at all. Any cluster-scoped object on the consumer side is copied 1:1 to the provider side.
 
 !!! warning
     Due to the obvious downsides of this approach, `none` should be used only in special circumstances.


xrstf:
@coderabbitai review!

docs/content/usage/synchronization.md
```diff
@@ -0,0 +1,173 @@
...
+!!! note
+    For natively namespaced resources (those that are namespaced in both the provider and consumer cluster), this is always the strategy being used by the konnector. The other isolation modes only influence how the konnector deals with cluster-scoped objects in the consumer cluster.
+
+### Cluster-Scoped
+
+Cluster-scoped objects require a different approach to isolation than namespaced objects. kube-bind offers three different so-called isolation strategies to deal with them. The strategy to use is configured globally via the `--cluster-scoped-isolation` CLI flag on the kube-bind backend and will from there affect all services offered by that backend.
+
+#### None Strategy
+
+The `none` strategy providers no consumer-separation at all. Any cluster-scoped object on the consumer side is copied 1:1 to the provider side.
```
cnvergence:
```suggestion
The `none` strategy provides no consumer-separation at all. Any cluster-scoped object on the consumer side is copied 1:1 to the provider side.
```

xrstf: You get one EagleEye point for spotting one of the three intentional typos!

PR #395, opened 2025-11-21
Add GatewayAPI support & docs

mjudeikis:
## Summary

Add docs how to deploy in "real world" and add GatewayAPI support vs old ingress. 
Dropping TLS support as gateway api default pattern is to terminate. 

## What Type of PR Is This?
/kind cleanup
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add GatewayAPI support
```

### DIFF
--- a/deploy/charts/backend/templates/deployment.yaml
+++ b/deploy/charts/backend/templates/deployment.yaml
@@ -82,15 +82,25 @@ spec:
             {{- if .Values.backend.cookieEncryptionKey }}
             - --cookie-encryption-key={{ .Values.backend.cookieEncryptionKey }}
             {{- end }}
+            {{- if .Values.backend.tls.enabled }}
+            - --tls-cert-file={{ .Values.backend.tls.tlsCertFile }}
+            - --tls-key-file={{ .Values.backend.tls.tlsKeyFile }}
+            {{- end }}
             {{- if eq .Values.backend.oidc.type "embedded" }}
             - --oidc-type=embedded
             {{- else if eq .Values.backend.oidc.type "external" }}
             - --oidc-type=external
             {{- end }}
           ports:
+            {{- if .Values.backend.tls.enabled }}
+            - name: https
+              containerPort: 8443
+              protocol: TCP
+            {{- else }}
             - name: http
               containerPort: 8080
               protocol: TCP
+            {{- end }}
           {{- with .Values.livenessProbe }}
           livenessProbe:
             {{- toYaml . | nindent 12 }}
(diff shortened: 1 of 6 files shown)


docs/content/setup/helm.md
```diff
@@ -41,21 +40,34 @@ The following prerequisites are required. Click the links below for detailed set
    - Update hostnames to match your setup
 
 3. **Install the backend using OCI chart:**
-   ```bash
+
+  Note !!!
+    To install production configuration, you will need to have OIDC provider. 
+    If you interested just to try - check quickstart [quickstart].
```
xrstf: `For more information, just check out the [quickstart guide].`

docs/content/setup/helm.md
```diff
@@ -41,21 +40,34 @@ The following prerequisites are required. Click the links below for detailed set
...
-       kube-bind oci://ghcr.io/kube-bind/charts/backend --version ${VERSION}
-
-   # Or install a specific development version
-   helm upgrade --install \
-       --namespace kube-bind \
-       --create-namespace \
-       --values ./deploy/charts/backend/examples/values-local-development.yaml \
-       kube-bind oci://ghcr.io/kube-bind/charts/backend --version 0.0.0-a50df39d7e4c71f7808f4209ec23f294c5ac8f86
-   ```
+helm upgrade --install \
```
xrstf: Let's put `--install` on its own line, like every other flag 😁

cnvergence (approved):

deploy/charts/backend/templates/deployment.yaml
```diff
@@ -82,25 +82,15 @@ spec:
             {{- if .Values.backend.cookieEncryptionKey }}
             - --cookie-encryption-key={{ .Values.backend.cookieEncryptionKey }}
             {{- end }}
-            {{- if .Values.backend.tls.enabled }}
```
xrstf, line 85: So TLS is simply not supported anymore when using Ingresses? Can you explain your thinking in removing the flags from the chart, but keeping all the TLS code in the backend?

mjudeikis: GatewayAPI has TLSRoute, where we would need to do Passthrouth, but its experimental channel (https://gateway-api.sigs.k8s.io/concepts/api-overview/) so it is not installed by default :/ feels like everybody is doing TLS termination these days. 

We could do TLS too, but this needs dragging in an experimental feature from the gateway api. And for now, I wanted to keep it simple.

mjudeikis: For now - lets keep it in the backend, if somebody would want to use it with this. You can still use it with ServiceType: Loadbalancer

xrstf: > if somebody would want to use it with this. You can still use it with ServiceType: Loadbalancer

That was kind of why I asked. Can't we keep all the tls.enabled-guarded YAML intact, for those who want to use the LoadBalancer instead?

mjudeikis: make sense. will update

mjudeikis: reverted those changes

docs/content/setup/helm.md
```diff
@@ -281,20 +304,4 @@ curl -s https://api.github.com/repos/kube-bind/kube-bind/releases | grep '"tag_n
...
-# Install latest stable release (recommended for production)
-helm upgrade --install kube-bind oci://ghcr.io/kube-bind/charts/backend --version ${VERSION}
-
-# Install specific release version
-helm upgrade --install kube-bind oci://ghcr.io/kube-bind/charts/backend --version 1.0.0
-
-# Install development build (for testing)
-helm upgrade --install kube-bind oci://ghcr.io/kube-bind/charts/backend --version 0.0.0-a1b2c3d
-```
+```
```
cnvergence, line 307: nit: leftover

mjudeikis: what left over?

cnvergence: nevermind :D

PR #405, opened 2025-11-27
Pre 0.6.0 release fixes

mjudeikis:
## What Type of PR Is This?
/kind bug

Start documentation for intergrations. Adding cert-manager for now.
Add some debug logging after deep deep debugging.
Add test for ^ 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -218,11 +218,7 @@ func IsClaimedWithReference(
 				}
 				logger.V(4).Info("listing objects for reference claim", "gvr", gvr)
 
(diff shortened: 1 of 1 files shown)


pkg/resources/resources.go
```diff
@@ -208,6 +211,13 @@ func IsClaimedWithReference(
 			defer cancel()
 
 			for _, version := range versions {
+				gvr := schema.GroupVersionResource{
```
ntnn, line 214: Use this GVR in L221 as well?

mjudeikis: yes :) done

PR #411, opened 2025-11-28
Re-check previously owned

mjudeikis:
## What Type of PR Is This?

If I create sync relationship using reference (Cert-Manager) and delete the object after - secret might be still lurking. For this, if we still have label indiciating that it was "claimed before"  - we need to re-check again. 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix dangling reference objects
```

### DIFF
--- a/pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
+++ b/pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
@@ -237,7 +237,7 @@ func (c *controller) enqueueConsumer(logger klog.Logger, obj interface{}) {
 	// If owner label is set - it was claimed before and might not be claimed now (example: reference gone or changed).
 	// So we need to enqueue it to check if it is still needed.
 	_, wasClaimedBefore := o.GetLabels()[kubebindv1alpha2.ObjectOwnerLabel]
-	if !c.isClaimed(logger, o, true) && !wasClaimedBefore {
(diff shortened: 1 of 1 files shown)


pkg/konnector/controllers/cluster/claimedresources/claimedresources_controller.go
```diff
@@ -234,7 +234,10 @@ func (c *controller) enqueueConsumer(logger klog.Logger, obj interface{}) {
 		return
 	}
 
-	if !c.isClaimed(logger, o, true) {
+	// If owner label is set - it was claimed before and might not be claimed now (example: reference gone or changed).
+	// So we need to enqueue it to check if it is still needed.
+	_, wasClaimedBefore := o.GetLabels()[kubebindv1alpha2.ObjectOwnerLabel]
+	if !c.isClaimed(logger, o, true) && !wasClaimedBefore {
```
ntnn: Doesn't this make more sense the other way around?
Not sure if the compiler optimizes that away but I'd put the "simpler" case first, so if the owner label is set the `.isClaimed` is skipped instead of first running `.isClaimed` and then checking the less intensive case

cnvergence: We could also check it inside isClaimed func, but as this helps to unblock dangling objects, we can revisit it

mjudeikis: make sense

PR #412, opened 2025-11-28
Scope down default deployment

mjudeikis:
## Summary

By default  deploy backend with lower informer scope to be "more secure by default" from the start.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/deploy/charts/backend/templates/deployment.yaml
+++ b/deploy/charts/backend/templates/deployment.yaml
@@ -76,8 +76,8 @@ spec:
             {{- if .Values.backend.consumerScope }}
             - --consumer-scope={{ .Values.backend.consumerScope }}
             {{- end }}
-            {{- if .Values.backend.informerScope }}
-            - --cluster-scoped-isolation={{ .Values.backend.informerScope }}
+            {{- if .Values.backend.clusterScopeIsolation }}
+            - --cluster-scoped-isolation={{ .Values.backend.clusterScopeIsolation }}
             {{- end }}
             {{- if .Values.backend.cookieSigningKey }}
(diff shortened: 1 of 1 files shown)


deploy/charts/backend/templates/deployment.yaml
```diff
@@ -76,6 +76,9 @@ spec:
             {{- if .Values.backend.consumerScope }}
             - --consumer-scope={{ .Values.backend.consumerScope }}
             {{- end }}
+            {{- if .Values.backend.informerScope }}
+            - --cluster-scoped-isolation={{ .Values.backend.informerScope }}
```
xrstf: This is the wrong flag or the wrong variable. The isolation mode has absolutely nothing to do with how the informer scopes are set up.

mjudeikis: yes, was wrong valyue.

deploy/charts/backend/templates/deployment.yaml
```diff
@@ -73,8 +73,11 @@ spec:
             {{- if .Values.backend.namespacePrefix }}
             - --namespace-prefix={{ .Values.backend.namespacePrefix }}
             {{- end }}
-            {{- if .Values.backend.consumerScope }}
-            - --consumer-scope={{ .Values.backend.consumerScope }}
+            {{- if .Values.backend.informerScope }}
+            - --consumer-scope={{ .Values.backend.informerScope }}
```
xrstf: According to the values.yaml, this field is called `consumerScope`, not `informerScope`.

PR #418, opened 2025-12-05
replace go-install with µget, cleanup Makefile

xrstf:
## Summary
This follows our changes in kcp and its related sibling projects. µget is superior to go-install because

* it can handle Go 1.25 compatibility hacks
* can and is optimized for installing non-Go binaries
* it does checksumming
* it can print the resulting binary path
* it can use a system-wide cache, great for when many projects use µget and all of them want golangci-lint
* it doesn't suffer from the NIH syndrome

Like in the other projects, the shell scripts themselves are kept/meant to be usable without make, so they install (using make, however) their dependencies as they see fit.

I also unfuddled with TOOLS_GO_BIN_DIR thingymabob, which seems to have been mostly broken and unusable. From now on, all binaries we produce end up in `bin/`, all tools we use from external sources are in `hack/tools`. It made no senses to me to place applyconfiguration-gen in the same folder as the backend binary.

I also took the liberty to move some of the larger chunks of shell code and moved them into actual script files. My brain is bad at reading and writing 20+ line inline bash scripts in Makefiles.

## What Type of PR Is This?
/kind cleanup

## Release Notes
```release-note
NONE
```


mjudeikis:
@coderabbitai review this please

PR #426, opened 2025-12-18
Standard errors & session store

mjudeikis:
## What Type of PR Is This?
/kind feature
/kind api-change

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Standardise UI/API errors 
Session verification 
Make embedded user static & readable 
```

### DIFF
--- a/backend/session/session.go
+++ b/backend/session/session.go
@@ -49,7 +49,7 @@ func (s *State) Encode() ([]byte, error) {
 
 // IsExpired checks if the session has expired
 func (s *State) IsExpired() bool {
-	return !s.ExpiresAt.IsZero() && time.Now().After(s.ExpiresAt)
+	return s.ExpiresAt.IsZero() || time.Now().After(s.ExpiresAt)
 }
 
 // SetExpiration sets the session expiration time


backend/auth/middleware.go
```diff
@@ -117,6 +148,50 @@ func (am *AuthMiddleware) AuthenticateRequest(next http.Handler) http.Handler {
...
+	})
+}
+
+// isValidSession checks if a session ID exists and hasn't expired
+func (am *AuthMiddleware) isValidSession(sessionID string) bool {
+	sessionInfo, err := am.sessionStore.Load(sessionID)
+	if err != nil {
+		return false
+	}
+	return time.Now().Before(sessionInfo.ExpiresAt)
```
gman0, line 191: What would be the consequences if `time.Now().Before(sessionInfo.ExpiresAt)` returns `true` at this very moment, but the request finishes already at `time.Now() > sessionInfo.ExpiresAt`? Not sure what the consensus is on these time-based checks. I believe I've seen some implementations with a buffer period where `should_renew_at > valid_since && should_renew_at < expires_at`. I'm just curious how this would be handled rather than saying it needs to change because I'm not sure what the consequences would be :)

The same is true for `State.IsExpired()`.

backend/auth/middleware.go
```diff
@@ -117,6 +148,50 @@ func (am *AuthMiddleware) AuthenticateRequest(next http.Handler) http.Handler {
...
+		case state.Token.Subject == "" || state.Token.Issuer == "" || state.SessionID == "":
+			logger.V(2).Info("Invalid session state: missing required fields")
+			writeErrorResponse(w, http.StatusUnauthorized, kubebindv1alpha2.ErrorCodeAuthenticationFailed, "Authentication required", "Invalid session state: missing required fields")
+			return
+		case state.IsExpired():
+			logger.V(2).Info("Session expired", "sessionID", state.SessionID)
+			writeErrorResponse(w, http.StatusUnauthorized, kubebindv1alpha2.ErrorCodeAuthenticationFailed, "Authentication required", "Session has expired")
+			return
+		case !am.isValidSession(state.SessionID):
+			logger.V(2).Info("Session ID not found or expired", "sessionID", state.SessionID)
```
gman0: Why check expiration twice?

backend/http/handler.go
```diff
@@ -48,6 +50,36 @@ var noCacheHeaders = map[string]string{
...
+	}
+}
+
+// mapErrorToCode maps common errors to structured error codes
+func mapErrorToCode(err error) (statusCode int, code string, details string) {
+	if apierrors.IsNotFound(err) {
+		return http.StatusNotFound, kubebindv1alpha2.ErrorCodeResourceNotFound, err.Error()
+	}
+	if apierrors.IsUnauthorized(err) {
+		return http.StatusUnauthorized, kubebindv1alpha2.ErrorCodeAuthenticationFailed, err.Error()
```
gman0, line 71: `IsUnathorized` and `ErrorCodeAuthenticationFailed`? Shouldn't it be `ErrorCodeAuthorizationFailed`?

backend/http/handler.go
```diff
@@ -252,21 +286,28 @@ func (h *handler) handleTemplates(w http.ResponseWriter, r *http.Request) {
 	prepareNoCache(w)
 	logger.Info("getting templates")
 
+	authenticated := auth.GetAuthContext(r.Context()).IsValid
+	if !authenticated {
+		logger.Info("unauthenticated request to get templates")
+		writeErrorResponse(w, http.StatusUnauthorized, kubebindv1alpha2.ErrorCodeAuthenticationFailed, "Authentication required", "Valid authentication credentials are required to access templates")
```
gman0, line 293: `!authenticated` and `http.StatusUnauthorized`? Wouldn't `http.Forbidden` be a better fit?

backend/session/store.go
```diff
@@ -0,0 +1,66 @@
...
+var ErrSessionNotFound = fmt.Errorf("session not found")
+
+type Store interface {
+	Save(state *State) error
+	Load(sessionID string) (*State, error)
+	Delete(sessionID string) error
+}
+
+type InMemoryStore struct {
+	lock     sync.Mutex
```
gman0: What access pattern do you foresee for this `InMemoryStore`? Would `RWMutex` make sense here?

docs/content/developers/backend/index.md
```diff
@@ -63,3 +63,17 @@ The backend includes several controllers for managing the export/binding lifecyc
...
+
+The backend currently uses in-memory session storage for OIDC authentication sessions. This creates the following limitation:
+
+- **Single Replica Constraint**: The backend cannot be run with replicas > 1 due to the lack of external session storage implementation
+- **Session Persistence**: Sessions are not persisted across backend restarts
+- **Load Balancer Issues**: If deployed behind a load balancer with multiple replicas, users may lose their sessions when requests are routed to different backend instances
+
+**Workaround**: Deploy the backend as a single replica (replicas: 1) until external session storage support is implemented.
+
+**Future Enhancement**: Implementation of external session storage (Redis, database, etc.) would resolve this limitation and enable high availability deployments.
```
gman0: Is there a ticket users could check to see the status of this future enhancement?

backend/session/store.go
```diff
@@ -0,0 +1,66 @@
...
+	sessions map[string]*State
+}
+
+func NewInMemoryStore() *InMemoryStore {
+	return &InMemoryStore{
+		lock:     sync.Mutex{},
```
gman0: Nit: `lock` is already default-constructed.

backend/auth/handler.go
```diff
@@ -282,22 +292,26 @@ func (ah *AuthHandler) createSessionState(authCode *AuthorizeRequest, token *oau
...
 			Subject: idToken.Subject,
 			Issuer:  idToken.Issuer,
+			Groups:  idToken.Groups,
 		},
 		SessionID:   authCode.SessionID,
 		ClusterID:   authCode.ClusterID,
 		RedirectURL: authCode.RedirectURL,
-	}, nil
+	}
+	s.SetExpiration(time.Hour)
```
gman0: Is this expiration value ever going to be user configurable?

sdk/apis/kubebind/v1alpha2/error_types.go
```diff
@@ -0,0 +1,67 @@
...
+// Error represents a structured error response from the kube-bind backend
+// +kubebuilder:object:root=true
+type Error struct {
+	metav1.TypeMeta `json:",inline"`
+	// Message is the human-readable error message
+	Message string `json:"message"`
+	// Code is the error code that can be used for programmatic handling
+	Code string `json:"code"`
+	// Details provides additional context about the error
+	Details string `json:"details,omitempty"`
```
gman0, line 33: `Details` has `,omitempty` tag but is not `+optional`.

backend/auth/middleware.go
```diff
@@ -49,19 +52,48 @@ type AuthContext struct {
 
 type AuthMiddleware struct {
 	jwtService          *JWTService
+	kubernetesMananger  *kubernetes.Manager
```
cnvergence:
```suggestion
	kubernetesManager  *kubernetes.Manager
```

backend/auth/middleware.go
```diff
@@ -49,19 +52,48 @@ type AuthContext struct {
...
+	if err := json.NewEncoder(w).Encode(errorResponse); err != nil {
+		// Fallback to plain text if JSON encoding fails
+		http.Error(w, message, statusCode)
+	}
+}
+
+func NewAuthMiddleware(
+	jwtService *JWTService,
+	cookieSigningKey, cookieEncryptionKey []byte,
+	kubernetesMananger *kubernetes.Manager,
```
cnvergence: same here

backend/auth/middleware.go
```diff
@@ -102,12 +133,12 @@ func (am *AuthMiddleware) AuthenticateRequest(next http.Handler) http.Handler {
 				s := securecookie.New(am.cookieSigningKey, am.cookieEncryptionKey)
 				state := &session.State{}
 				if err := s.Decode(cookieName, cookie.Value, state); err == nil {
+					// Only set as valid after all checks pass
 					authCtx.SessionState = state
-					authCtx.IsValid = true
 					authCtx.ClientType = ClientTypeUI
-				} else {
-					logger.V(2).Info("Failed to decode session cookie", "error", err)
 				}
```
cnvergence, line 142: additional brace here?

backend/oidc/oidc.go
```diff
@@ -47,15 +47,28 @@ func New(caBundleFile string, listener net.Listener, addrOverride string) (*Serv
 		s.tlsConfig = tlsConfig
 	}
 
+	statisUserProvider := NewStaticUserProvider()
```
cnvergence:
```suggestion
	staticUserProvider := NewStaticUserProvider()
```

backend/session/session.go
```diff
@@ -31,13 +33,27 @@ type State struct {
...
 	ClusterID   string    `msgpack:"cid,omitempty"`
 	RedirectURL string    `msgpack:"red,omitempty"`
+	CreatedAt   time.Time `msgpack:"cat,omitempty"`
+	ExpiresAt   time.Time `msgpack:"eat,omitempty"`
 }
 
 type TokenInfo struct {
-	Subject string `msgpack:"sub,omitempty"`
-	Issuer  string `msgpack:"iss,omitempty"`
+	Subject string   `msgpack:"sub,omitempty"`
```
cnvergence, line 41: should we validate the session?

backend/auth/middleware.go
mjudeikis: Basically, it would load the page and still binding completes/allowed. But we need to implement renewal flow. 
I think its not such a bit problem. bigger problem is somebody using 2 weeks old token :)

backend/auth/middleware.go
mjudeikis: One checks if the session in the cookie has expired and just denies it.
Second checks if the same session has expired in the session store. So if one would want to revoke a session, it just needs to set it to expired in the session store (future future), so even if the user has a valid session, it still can be revoked.

backend/http/handler.go
mjudeikis: I think (if I recall my thinking behind this) - its bit of kubernetizm:
IsUnauthorized - lack of authentication
isForbidden - no permission hence no authorization.

Or Im reading this wrong?

backend/http/handler.go
mjudeikis: I dont think so. Forbiden is when you are Authenticated but dont have permissions to access resources

backend/session/store.go
mjudeikis: It would. changed

backend/auth/handler.go
mjudeikis: Added flag.

sdk/apis/kubebind/v1alpha2/error_types.go
mjudeikis: This was not used in k8s machienry so kinda forgot :)

backend/auth/middleware.go
mjudeikis: why?

backend/session/session.go
mjudeikis: wdym?

cnvergence: nvm, I see it happens in the middleware

backend/auth/middleware.go
cnvergence: I mean, if you have changed it intentionally, we could just add/change an error message if the cookie was not found.

backend/auth/middleware.go
gman0: Ah, the renewal flow is not yet in place? Okay, probably not much more we can do here at the moment. Thanks for explaining!

backend/auth/middleware.go
mjudeikis: changed this a bit. but still not sure what that comment means :)

docs/content/developers/backend/index.md
mjudeikis: added

backend/auth/middleware.go
cnvergence: looks good now, thanks :D

backend/http/handler.go
gman0: Yes all good, my bad. The check for Forbidden actually checks for authorization failure.

gman0 (dismissed):
/lgtm

backend/session/session.go
```diff
@@ -31,13 +33,27 @@ type State struct {
...
+	Groups  []string `msgpack:"grp,omitempty"`
 }
 
 func (s *State) Encode() ([]byte, error) {
 	return msgpack.Marshal(s)
 }
+
+// IsExpired checks if the session has expired
+func (s *State) IsExpired() bool {
+	return !s.ExpiresAt.IsZero() && time.Now().After(s.ExpiresAt)
```
ntnn:
```suggestion
	return s.ExpiresAt.IsZero() || time.Now().After(s.ExpiresAt)
```
Should this be this? The expiration should never be zero. The current code would yield not expired if expired at is zero.

mjudeikis: One could set expired at 0 to make sure its never expires? 
Depends if we want to have this here or not

gman0: It would be nice to have a unit test for all these different cases.

PR #428, opened 2025-12-22
Add authz & cluster anchor

mjudeikis:
## What Type of PR Is This?

Add object `Kind: Cluster` in the provider cluster to anchor for reconciler-based configuration
Add authz permissions (bind)
Add allowedUsers, allowedGroups
Add default configuration for embedded user 

/kind api-change

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add bind permission & enhance rbac security for individual clusters
```

### DIFF
--- a/backend/kubernetes/manager.go
+++ b/backend/kubernetes/manager.go
@@ -48,7 +48,8 @@ type Manager struct {
 	externalCA               []byte
 	externalTLSServerName    string
 
-	manager mcmanager.Manager
+	manager      mcmanager.Manager
+	embeddedOIDC bool
 }
 
 func NewKubernetesManager(
@@ -59,6 +60,7 @@ func NewKubernetesManager(
 	externalCA []byte,
 	externalTLSServerName string,
 	manager mcmanager.Manager,
+	embeddedOIDC bool,
 ) (*Manager, error) {
 	m := &Manager{
 		namespacePrefix:    namespacePrefix,
@@ -69,7 +71,8 @@ func NewKubernetesManager(
 		externalCA:               externalCA,
 		externalTLSServerName:    externalTLSServerName,
 
-		manager: manager,
+		manager:      manager,
+		embeddedOIDC: embeddedOIDC,
 	}
 
 	if err := m.manager.GetFieldIndexer().IndexField(ctx, &corev1.Namespace{}, NamespacesByIdentity,
@@ -259,9 +262,11 @@ func (m *Manager) AuthorizeRequest(ctx context.Context, subject string, groups [
 		return err
 	}
 
-	groups = append(groups, "system:authenticated")
+	if m.embeddedOIDC {
+		groups = append(groups, "system:authenticated")
+	}
 
-	// Check if user has access to create pods (basic permission test)
+	// Check if user can bind (basic permission test)
 	sar := &authzv1.SubjectAccessReview{
 		Spec: authzv1.SubjectAccessReviewSpec{
 			User:   subject,
@@ -279,10 +284,7 @@ func (m *Manager) AuthorizeRequest(ctx context.Context, subject string, groups [
 		return err
 	}
 
-	if sarResponse.Status.Allowed {
-		logger.Info("User is allowed to bind", "user", subject)
-		// Proceed with your controller logic
-	} else {
+	if !sarResponse.Status.Allowed {
 		logger.Info("User is not allowed to bind", "user", subject, "reason", sarResponse.Status)
 		// Return a structured authorization error
 		return errors.NewForbidden(
@@ -291,6 +293,7 @@ func (m *Manager) AuthorizeRequest(ctx context.Context, subject string, groups [
 			fmt.Errorf("user %q is not authorized to bind resources: %s", subject, sarResponse.Status.Reason),
 		)
 	}
+	logger.Info("User is allowed to bind", "user", subject)
 	return nil
 }
 
--- a/backend/provider/kcp/provider.go
+++ b/backend/provider/kcp/provider.go
@@ -79,8 +79,11 @@ type awareWrapper struct {
 }
 
 func (a *awareWrapper) Engage(ctx context.Context, name string, cluster cluster.Cluster) error {
+	ctx, cancel := context.WithCancel(ctx) //nolint:govet // cancel is called in the error case only.
+
 	err := a.Aware.Engage(ctx, name, cluster)
 	if err != nil {
+		cancel()
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


backend/controllers/cluster/cluster_controller.go
```diff
@@ -0,0 +1,122 @@
...
+		logger.Info("Cluster status updated")
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ClusterReconciler) SetupWithManager(mgr mcmanager.Manager) error {
+	return mcbuilder.ControllerManagedBy(mgr).
+		For(&kubebindv1alpha2.Cluster{}).
```
ntnn, line 115: Do we expect consumers to also have the Cluster CRD installed?

backend/kubernetes/manager.go
```diff
@@ -240,3 +243,71 @@ func (m *Manager) ListDynamicResources(ctx context.Context, cluster string, gvk
...
+
+	// Create the authorization clientset
+	authClient, err := authorizationv1.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+
+	groups = append(groups, "system:authenticated")
+
+	// Check if user has access to create pods (basic permission test)
```
ntnn:
```suggestion
	// Check if user can bind (basic permission test)
```

backend/kubernetes/manager.go
```diff
@@ -240,3 +243,71 @@ func (m *Manager) ListDynamicResources(ctx context.Context, cluster string, gvk
...
+	} else {
+		logger.Info("User is not allowed to bind", "user", subject, "reason", sarResponse.Status)
+		// Return a structured authorization error
+		return errors.NewForbidden(
+			schema.GroupResource{Group: "kube-bind.io", Resource: "bind"},
+			"",
+			fmt.Errorf("user %q is not authorized to bind resources: %s", subject, sarResponse.Status.Reason),
+		)
+	}
+	return nil
```
ntnn:
```suggestion
	if !sarResponse.Status.Allowed {
		logger.Info("User is not allowed to bind", "user", subject, "reason", sarResponse.Status)
		// Return a structured authorization error
		return errors.NewForbidden(
			schema.GroupResource{Group: "kube-bind.io", Resource: "bind"},
			"",
			fmt.Errorf("user %q is not authorized to bind resources: %s", subject, sarResponse.Status.Reason),
		)
	}
	logger.Info("User is allowed to bind", "user", subject)
	return nil
```

Just reordering a bit. I think code reads easier if error cases check and return early.

backend/options/oidc.go
```diff
@@ -75,6 +85,8 @@ func (options *OIDC) Complete(listener net.Listener) error {
 		options.IssuerClientSecret = cfg.ClientSecret
 		// This should be provided from outside, but in embedded - we detect.
 		options.CallbackURL = cfg.CallbackURL
+
+		options.AllowedGroups = append(options.AllowedGroups, "system:authenticated")
```
ntnn, line 89: Missing `.AllowedUsers`?

backend/provider/kcp/provider.go
```diff
@@ -73,6 +79,43 @@ type awareWrapper struct {
 }
 
 func (a *awareWrapper) Engage(ctx context.Context, name string, cluster cluster.Cluster) error {
-	// TODO: Add cluster seeding here in the follow-up PRS.
-	return a.Aware.Engage(ctx, name, cluster)
+	err := a.Aware.Engage(ctx, name, cluster)
+	if err != nil {
+		return err
+	}
```
ntnn, line 88: I think the context should be wrapped, so when the `Cluster` fails to be created below the context can be cancelled again. Otherwise if the same cluster is engaged again this engage will error because the old cluster is still engaged but no Cluster object was created.

backend/controllers/cluster/cluster_controller.go
mjudeikis: No, just provider

mjudeikis: No, just provider

ntnn: Then we'll need filtering here or only watch the local/provider cluster with this controller. I think it'd be a good idea to setup the e2e tests to only install the CRDs in the respective clusters that we expect. 
I wouldn't let that block here now but we should make an issue to track that before we push this out to consumers.

mjudeikis: This is backend only, this always will be only the backend cluster's.
I dont think we have this problem as we have dedicated PID process for consumer and provider

backend/options/oidc.go
mjudeikis: No. We use them "as is". We added authenticated as default group here and in auth. 
I dont think we need others. Gonna test manually this.

backend/provider/kcp/provider.go
mjudeikis: added. good call.

backend/options/oidc.go
mjudeikis: Ok, this path is only for embedded oidc/dev one. But there was bug around this. manual testing got it. Thanks.

backend/controllers/cluster/cluster_controller.go
mjudeikis: if ever we consolidate backend and konnector code -yes, for now - no need

ntnn: Ahh okay, yeah then no need

backend/options/oidc.go
ntnn: I also wasn't looking correctly :D I read `options.AllowedGroups = append(cfg.AllowedGroups, ...` and thought `.AllowedUsers` was missing.

backend/provider/kcp/provider.go
```diff
@@ -73,6 +79,48 @@ type awareWrapper struct {
...
+	const maxRetries = 3
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		err = cl.Create(ctx, obj)
+		if err == nil || apierrors.IsAlreadyExists(err) {
+			break
+		}
+
+		if !apierrors.IsForbidden(err) {
+			return err
+		}
```
ntnn, line 109:
```suggestion
		if !apierrors.IsForbidden(err) {
			cancel()
			return err
		}
```

mjudeikis: wups :)

backend/options/oidc.go
mjudeikis: yap. but double look and I found another bug so good :D

ntnn (approved):
/approve
/lgtm

PR #429, opened 2025-12-23
Add cluster-identity

mjudeikis:
## Summary

Re-add cluster-identity as a binding request. This would allow pre-provisioning of right identity namespaces based on cluster identity, rather than user identity (as it is now). 

This is a temporary solution until we can sort this out in a more sustainable way. Its bit hacky. 

## What Type of PR Is This?
/kind regression
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Re-add cluster identity 
```

### DIFF
--- a/backend/kubernetes/resources/namespace_test.go
+++ b/backend/kubernetes/resources/namespace_test.go
@@ -18,47 +18,52 @@ package resources
 
 import (
 	"context"
+	"crypto/sha256"
+	"fmt"
 	"strings"
 	"testing"
 
+	"github.com/martinlindhe/base36"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
+func expectedIdentityHash(identity string) string {
+	hash := sha256.Sum224([]byte(identity))
+	return strings.ToLower(base36.EncodeBytes(hash[:8]))
+}
+
 func TestCreateNamespace(t *testing.T) {
 	scheme := runtime.NewScheme()
 	_ = corev1.AddToScheme(scheme)
 
 	tests := []struct {
-		name              string
-		generateName      string
-		identity          string
-		author            string
-		wantErr           bool
-		wantNameGenerated bool
-		wantAnnotations   map[string]string
+		name            string
+		generateName    string
+		identity        string
+		author          string
+		wantErr         bool
+		wantAnnotations map[string]string
 	}{
 		{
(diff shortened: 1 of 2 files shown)


cli/pkg/kubectl/bind-apiservice/cmd/cmd.go
```diff
@@ -82,7 +82,11 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
 				fmt.Fprintf(streams.ErrOut, "%s\n\n", yellow("DISCLAIMER: This is a prototype. It will change in incompatible ways at any time."))
 			}
 
-			return opts.Run(cmd.Context())
+			if err := opts.Run(cmd.Context()); err != nil {
```
mjudeikis, line 85: This as causing double errors in the CLI

sdk/apis/kubebind/v1alpha2/bindingresponse_types.go
```diff
@@ -86,7 +86,15 @@ type BindableResourcesRequest struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
 
+	// TemplateRef specifies the APIServiceExportTemplate to bind to.
+	// +required
+	// +kubebuilder:validation:Required
 	TemplateRef APIServiceExportTemplateRef `json:"templateRef"`
+	// ClusterIdentity contains information that uniquely identifies the cluster.
+	// When doing dry run, we expect the client to fill this field in (or it will be taked from local cluster where context is available).
```
ntnn:
```suggestion
	// When doing dry run, we expect the client to fill this field in (or it will be taken from local cluster where context is available).
```

backend/kubernetes/resources/namespace_test.go
```diff
@@ -36,29 +34,34 @@ func TestCreateNamespace(t *testing.T) {
 	tests := []struct {
 		name              string
 		generateName      string
-		id                string
+		identity          string
+		author            string
```
ntnn: Author is the user and identity is the cluster identity?

mjudeikis: yes.

ntnn (dismissed):

mjudeikis:
> So we are creating one namespace per author/"consumer" to store the cluster objects? I wonder if it wouldn't be safer to make the namespace deterministic instead of using generateName, e.g. `kube-clusters-{identity-hash}`; that would prevent things like one identity "owning" multiple namespaces.
> 
> Edit: I wouldn't change the namespace handling in this PR. We can adjust that later; just came to mind while reading the code.

I think we can make this with this change too. Its kinda easy and either way - small breaking change.

ntnn (approved):

PR #432, opened 2025-12-30
Extend none isolation to APIServiceNamespaces

mjudeikis:
## What Type of PR Is This?

This prs renames `cluster-scope-isolation` to `isolation` so its broarder term
As `APIServiceNamespace` is key object to map namespaces, make it isolation aware. So now, if you use isolation none, all lands end up in the same namespace with the same names. Meaning we can actually get a twin situation. 

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
deprecate --cluster-scope-isolation in favor of --isolation
Extend isolation strategy None to not use prefixes and just copy as-is allowing twin type of configurations
```


backend/controllers/serviceexportrequest/serviceexportrequest_controller.go
```diff
@@ -49,9 +49,9 @@ type APIServiceExportRequestReconciler struct {
 	manager mcmanager.Manager
 	opts    controller.TypedOptions[mcreconcile.Request]
 
-	informerScope          kubebindv1alpha2.InformerScope
-	clusterScopedIsolation kubebindv1alpha2.Isolation
```
cnvergence, line 53: finally 👍🏻

backend/options/options.go
```diff
@@ -151,7 +151,10 @@ func (options *Options) AddFlags(fs *pflag.FlagSet) {
...
+	// TODO(mjudeikis): remove deprecated flag in future release
+	fs.StringVar(&options.Isolation, "cluster-scoped-isolation", options.Isolation, "How cluster scoped service objects are isolated between multiple consumers on the provider side. Among the choices, \"prefixed\" prepends the name of the cluster namespace to an object's name; \"namespaced\" maps a consumer side object into a namespaced object inside the corresponding cluster namespace; \"none\" is used for the case of a dedicated provider where isolation is not necessary.")
+	_ = fs.MarkDeprecated("cluster-scoped-isolation", "use --isolation instead")
+	fs.StringVar(&options.Isolation, "isolation", options.Isolation, "Deprecated: use --cluster-scoped-isolation instead. How cluster scoped service objects are isolated between multiple consumers on the provider side. Among the choices, \"prefixed\" prepends the name of the cluster namespace to an object's name; \"namespaced\" maps a consumer side object into a namespaced object inside the corresponding cluster namespace; \"none\" is used for the case of a dedicated provider where isolation is not necessary.")
```
cnvergence: shouldn't it be `use --isolation instead`?

sdk/apis/kubebind/v1alpha2/apiserviceexport_types.go
```diff
@@ -111,7 +111,17 @@ type APIServiceExportSpec struct {
 
 	// ClusterScopedIsolation specifies how cluster scoped service objects are isolated between multiple consumers on the provider side.
 	// It can be "Prefixed", "Namespaced", or "None".
+	// +deprecated
+	// Deprecated: use Isolation instead.
```
xrstf: IIRC to make linters understand this, there needs to be an empty comment line in the comment block before the "Deprecated:" line.

sdk/apis/kubebind/v1alpha2/apiserviceexport_types.go
```diff
@@ -111,7 +111,17 @@ type APIServiceExportSpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="clusterScopedIsolation is immutable"
 	ClusterScopedIsolation Isolation `json:"clusterScopedIsolation,omitempty"`
+
+	// Isolation specifies how service objects are isolated between multiple consumers on the provider side.
+	// It can be "Prefixed", "Namespaced", or "None".
```
xrstf: Personally I would leave this out, the ENUM makes it clear already, is machine-readable and does not potentially get out-of-sync with the code.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -267,16 +267,20 @@ func (r *reconciler) ensureControllerForSchema(ctx context.Context, export *kube
...
 		isolationStrategy = isolation.NewPrefixed(r.providerNamespace, providerNamespaceUID)
 
-	case export.Spec.ClusterScopedIsolation == kubebindv1alpha2.IsolationNamespaced:
+	case export.Spec.Isolation == kubebindv1alpha2.IsolationNamespaced:
+		logger.V(4).Info("Using Namespaced isolation strategy", "export", export.Name)
 		isolationStrategy = isolation.NewNamespaced(r.providerNamespace)
 	default:
 		// Default to None isolation strategy if no valid isolation strategy is specified
+		logger.V(4).Info("Using default None isolation strategy", "export", export.Name)
 		isolationStrategy = isolation.NewNone(r.providerNamespace, providerNamespaceUID)
```
xrstf, line 285: Food for thought, should this be a reconcile error? Is this just hiding broken objects from getting noticed by the admin?

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
```diff
@@ -144,7 +144,7 @@ func (r *reconciler) ensureControllers(ctx context.Context, namespace, name stri
 		}
 
 		processedSchemas[name] = true // This is only schemas names (suffix)
-		isClusterScoped = schema.Spec.Scope == apiextensionsv1.ClusterScoped || schema.Spec.InformerScope == kubebindv1alpha2.ClusterScope
+		isClusterScoped = schema.Spec.InformerScope == kubebindv1alpha2.ClusterScope
```
xrstf, line 147: Is this something we should mention in the release-notes? It seems like kube-bind is now doing less automagicalism to determine the correct scope to use, and instead the admin now needs to make sure their InformerScopes are set correctly? Or was the automagicalism already not doing anything? My brain currently can't parse all combinations ^^

docs/content/developers/architecture.md
```diff
@@ -0,0 +1,315 @@
+# Architecture Overview
```
xrstf, line 1: Large parts of this duplicate what we already have on https://docs.kube-bind.io/main/usage/synchronization/#cluster-isolation -- can we at least link to that doc for folks who accidentally stumbled into the dev section and are just looking for a quicker overview?

cli/pkg/kubectl/dev/plugin/create.go
```diff
@@ -67,17 +70,32 @@ type DevOptions struct {
...
 }
 
+// assetVersion is the version of the kube-bind backend assets used in dev mode
+var assetVersion = ""
+
+// fallbackAssetVersion is used when unable to fetch the latest version
+var fallbackAssetVersion = "0.6.0"
+
+// GitHubRelease represents a GitHub release response
+type GitHubRelease struct {
```
xrstf: This can be a private struct.

cli/pkg/kubectl/dev/plugin/create.go
```diff
@@ -67,17 +70,32 @@ type DevOptions struct {
...
+	TagName string `json:"tag_name"`
+}
+
 // NewDevOptions creates a new DevOptions
 func NewDevOptions(streams genericclioptions.IOStreams) *DevOptions {
 	opts := base.NewOptions(streams)
+	// Initialize assetVersion with fallback if not set
```
xrstf: This comment is not adding much :-)

cli/pkg/kubectl/dev/plugin/create.go
```diff
@@ -92,15 +110,74 @@ func (o *DevOptions) AddCmdFlags(cmd *cobra.Command) {
...
+	var release GitHubRelease
+	if err := json.Unmarshal(body, &release); err != nil {
+		return "", fmt.Errorf("failed to parse release data: %w", err)
+	}
+
+	if release.TagName == "" {
+		return "", fmt.Errorf("no tag name in release data")
+	}
+
+	// Remove 'v' prefix if present
```
xrstf: This comment is not adding much :-)

cli/pkg/kubectl/dev/plugin/create.go
```diff
@@ -67,17 +70,32 @@ type DevOptions struct {
 	KindNetwork         string
 }
 
+// assetVersion is the version of the kube-bind backend assets used in dev mode
+var assetVersion = ""
+
+// fallbackAssetVersion is used when unable to fetch the latest version
+var fallbackAssetVersion = "0.6.0"
```
xrstf: Let's make it a const.

cli/pkg/kubectl/dev/plugin/create.go
```diff
@@ -67,17 +70,32 @@ type DevOptions struct {
 	KindNetwork         string
 }
 
+// assetVersion is the version of the kube-bind backend assets used in dev mode
+var assetVersion = ""
```
xrstf: I do not understand how this variable is used. It's globale state and it gets defaulted very early to fallbackAssetVersion. Could this ever not be empty and then be defaulted?

backend/oidc/oidc_test.go
```diff
@@ -53,6 +53,7 @@ func TestLoadTLSConfig_Success(t *testing.T) {
 	// Verify the TLS config was created
 	if config == nil {
 		t.Fatal("Expected non-nil TLS config")
+		return
```
xrstf, line 56: Fatal() already ends the test early, no need for a return.

sdk/apis/kubebind/v1alpha2/apiserviceexport_types.go
mjudeikis: Leave you what? not sure I get this comment

mjudeikis: ah, got it

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
mjudeikis: This is required in the object, so it should never be unset. Will make it logging V2 so its visible in the logs for now,.

pkg/konnector/controllers/cluster/serviceexport/serviceexport_reconcile.go
mjudeikis: I think this was somehow leaking from old implementation. As much as I understand this didnt change the behaviour.

docs/content/developers/architecture.md
mjudeikis: added

cli/pkg/kubectl/dev/plugin/create.go
mjudeikis: reworked this.

backend/oidc/oidc_test.go
mjudeikis: yes, but linter does not like it :) something about return early to avoid NPE :D

cli/pkg/kubectl/dev/plugin/create.go
mjudeikis: done

cli/pkg/kubectl/dev/plugin/create.go
mjudeikis: reworked. was left over from iterration on this

backend/options/options.go
```diff
@@ -151,7 +151,10 @@ func (options *Options) AddFlags(fs *pflag.FlagSet) {
...
+	// TODO(mjudeikis): remove deprecated flag in future release
+	fs.StringVar(&options.Isolation, "cluster-scoped-isolation", options.Isolation, "How cluster scoped service objects are isolated between multiple consumers on the provider side. Among the choices, \"prefixed\" prepends the name of the cluster namespace to an object's name; \"namespaced\" maps a consumer side object into a namespaced object inside the corresponding cluster namespace; \"none\" is used for the case of a dedicated provider where isolation is not necessary.")
+	_ = fs.MarkDeprecated("cluster-scoped-isolation", "use --isolation instead")
+	fs.StringVar(&options.Isolation, "isolation", options.Isolation, "Deprecated: use --isolation instead. How cluster scoped service objects are isolated between multiple consumers on the provider side. Among the choices, \"prefixed\" prepends the name of the cluster namespace to an object's name; \"namespaced\" maps a consumer side object into a namespaced object inside the corresponding cluster namespace; \"none\" is used for the case of a dedicated provider where isolation is not necessary.")
```
xrstf: > fs.StringVar(&options.Isolation, "isolation", options.Isolation, "Deprecated: use --isolation instead.

So the --isolation flag says it is deprecated and the users should instead use --isolation?

mjudeikis: I think I went autopilot on this. https://github.com/kube-bind/kube-bind/pull/432#discussion_r2661437136 fixed now

xrstf (approved):
sip it

PR #445, opened 2026-01-22
Backend only mode

mjudeikis:
## Summary

This is backend only mode where we dont use oidc, web app and less steps to bind things. 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add backend only mode. 
```

### DIFF
--- a/backend/server.go
+++ b/backend/server.go
@@ -328,11 +328,10 @@ func (s *Server) Run(ctx context.Context) error {
 
 	if !s.Config.Options.FrontendDisabled {
 		return s.WebServer.Start(ctx)
-	} else {
-		logger.Info("Frontend is disabled; skipping web server start")
-		<-ctx.Done()
-		return nil
 	}
+	logger.Info("Frontend is disabled; skipping web server start")
+	<-ctx.Done()
+	return nil
 }
 
 func (s *Server) seedCluster(ctx context.Context) error {
--- a/sdk/apis/kubebind/v1alpha2/register.go
+++ b/sdk/apis/kubebind/v1alpha2/register.go
@@ -33,22 +33,6 @@ const (
 
 	// GroupVersion is the group version used in this package.
 	GroupVersion = "v1alpha2"
-
-	// Kind constants for the types in this package.
-	KindBoundSchema              = "BoundSchema"
-	KindAPIServiceBinding        = "APIServiceBinding"
-	KindAPIServiceBindingBundle  = "APIServiceBindingBundle"
-	KindAPIServiceExport         = "APIServiceExport"
-	KindAPIServiceExportRequest  = "APIServiceExportRequest"
-	KindAPIServiceNamespace      = "APIServiceNamespace"
-	KindClusterBinding           = "ClusterBinding"
-	KindBindingProvider          = "BindingProvider"
-	KindBindingResourceResponse  = "BindingResourceResponse"
-	KindAPIServiceExportTemplate = "APIServiceExportTemplate"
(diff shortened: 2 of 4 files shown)


backend/server.go
```diff
@@ -298,11 +326,13 @@ func (s *Server) Run(ctx context.Context) error {
...
+		return s.WebServer.Start(ctx)
+	} else {
+		logger.Info("Frontend is disabled; skipping web server start")
 		<-ctx.Done()
-		logger.Info("Context done")
-	}()
-	return s.WebServer.Start(ctx)
+		return nil
+	}
 }
```
ntnn, line 335:
```suggestion
	if !s.Config.Options.FrontendDisabled {
		return s.WebServer.Start(ctx)
	}

	logger.Info("Frontend is disabled; skipping web server start")
	<-ctx.Done()
	return nil
}
```

docs/proposals/backend-only-bindings.md
```diff
@@ -0,0 +1,114 @@
+# Backend-Only Bindings
```
ntnn, line 1: Do we need a proposal section here?

pkg/konnector/controllers/servicebindingbundle/servicebindingbundle_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	logger.Info("Starting controller")
+	defer logger.Info("Shutting down controller")
+
+	for i := 0; i < numThreads; i++ {
+		go wait.UntilWithContext(ctx, c.startWorker, time.Second)
+	}
+
+	// Start polling goroutine that re-queues all bundles every x seconds
+	// TODO(mjudeikis): make this a long lived watch on the provider cluster instead.
+	go c.startPoller(ctx)
```
ntnn, line 210: Why is this polling instead of a handler on an informer?

sdk/apis/kubebind/v1alpha2/register.go
```diff
@@ -33,6 +33,22 @@ const (
...
+	KindAPIServiceExportRequest  = "APIServiceExportRequest"
+	KindAPIServiceNamespace      = "APIServiceNamespace"
+	KindClusterBinding           = "ClusterBinding"
+	KindBindingProvider          = "BindingProvider"
+	KindBindingResourceResponse  = "BindingResourceResponse"
+	KindAPIServiceExportTemplate = "APIServiceExportTemplate"
+	KindCollection               = "Collection"
+	KindCluster                  = "Cluster"
+	KindBindableResourcesRequest = "BindableResourcesRequest"
+	KindAPIConversion            = "APIConversion"
```
ntnn: What are these useful for? I remember having used similar constants in the past but I don't really see where it's useful atm

xrstf: Same, I also think these types of constants add nothing of value in general.

docs/proposals/backend-only-bindings.md
mjudeikis: 🤷 didnt wanted to drop this document to /dev/null. But not sure if we need it into web docs.... maybe if we have more in the longer run

pkg/konnector/controllers/servicebindingbundle/servicebindingbundle_controller.go
mjudeikis: will create follow-up issue for this. just amount of work needed for this.

PR #451, opened 2026-01-28
Reworked RBAC machinery for APIServiceExports

### LINKED ISSUE #344: Implement ClusterScoped resource permissionClaims RBAC
When setting RBAC for permission claims, we act on APIServiceNamespace. For cluster scoped resources this does not work. We need to implement separate rbac machinery for this.

Look for this issue link in the code.

gman0:
## Summary

Added new APIServiceExportRBAC controller for managing APIServiceExport RBACs.

* All (Cluster)RoleBindings use the same subject: the `<APIServiceExport.Namespace>/kube-binder` ServiceAccount.
* The main ClusterRole `kube-binder-exports` aggregates all ClusterRoles of all service exports.
* Each APIServiceExport has its own set of ClusterRoles:
  * `kube-binder-resources-<APIServiceExport.Namespace>-<APIServiceExport.Name>` ClusterRole for its exported resources,
  * and `kube-binder-claims-<APIServiceExport.Namespace>-<APIServiceExport.Name>` ClusterRole for its claimed resources.
    * If the informer scope is Cluster: this ClusterRole contains rules for all export's claims.
    * If the informer scope is Namespaced: this ClusterRole contains rules only for cluster-scoped claimed resources.
      * For namespace-scoped claims, a `kube-binder-claims-<APIServiceExport.Name>` is created in each respective `<APIServiceNamespace.Status.Namespace>` namespace.

Additionally:
* `kube-binder` ClusterRoleBinding creation/update was moved to /bind handler
* Redundant boundschemas and boundschemas/status rules in `kube-binder-<ClusterBinding.Namespace>` ClusterRole were removed, because they are already present in `kube-binder` ClusterRole.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #344 

## Release Notes

```release-note
Reworked RBAC machinery for APIServiceExports
```

### DIFF
--- a/docs/content/developers/architecture.md
+++ b/docs/content/developers/architecture.md
@@ -278,6 +278,41 @@ Running with 2 consumers validates:
 - For cluster-scoped resources with IsolationPrefixed, resources are name-prefixed
 - For cluster-scoped resources with IsolationNamespaced, provider CRD is toggled to NamespaceScoped
 
+## RBAC Infrastructure
+
+### Exported resources
+
+Exported resources always use cluster-scoped RBACs:
+
+- Aggregating `kube-binder-exports` ClusterRole
+  - Bound with `kube-binder-exports` ClusterRoleBinding to `<APIServiceExport.Namespace>/kube-binder` ServiceAccount
+
+This aggregating ClusterRole includes all subsequent ClusterRoles that are created for each APIServiceExport.
+
+- `kube-binder-<APIServiceExport.Namespace>-<APIServiceExport.Name>` ClusterRole for each APIServiceExport
+  - Each resource in the APIServiceExport is granted `"get", "list", "watch", "create", "update", "patch", "delete"` set of verbs.
+  - Aggregates into `kube-binder-exports` ClusterRole.
+
+### Claimed resources
+
+Claimed resources use namespace- or cluster-scoped RBACs depending on the resource scope of the claimed resource itself, as well as backend's informer scope.
+
+For cluster-scoped informers:
+
+**Scope of the claimed resource** | **Informer scope** | **What happens**
+--- | --- | ---
+cluster | cluster | Creates ClusterRole & ClusterRoleBinding
+namespace | cluster | Creates ClusterRole & ClusterRoleBinding
+
+For namespace-scoped informers:
+
+**Scope of the claimed resource** | **Informer scope** | **What happens**
+--- | --- | ---
+cluster | namespace | Creates ClusterRole & ClusterRoleBinding
+namespace | namespace | Creates Role & RoleBinding
+
+Regardless of the configuration, claimed resources are granted `"get", "list", "watch", "create", "update", "patch", "delete"` set of verbs.
+
 ## Implementation Details
 
 ### Code Structure


backend/controllers/clusterbinding/clusterbinding_reconcile.go
```diff
@@ -120,187 +82,3 @@ func (r *reconciler) ensureClusterBindingConditions(_ context.Context, clusterBi
...
-					UID:        ns.UID,
-				},
-			},
-		},
-		Rules: []rbacv1.PolicyRule{
-			// Always need to be able to get/list/watch the BoundSchemas
-			// to be able to figure out what to bind.
-			{
-				APIGroups: []string{kubebindv1alpha2.GroupName},
-				Resources: []string{"boundschemas"},
```
gman0, line 158: ^ These rules are already part of the kube-binder ClusterRole:

https://github.com/kube-bind/kube-bind/blob/d15b44c34d7d5844482cdb59ff6256912acc9c69/backend/kubernetes/resources/rbac.go#L105-L114

Probably no need to re-create them?

backend/kubernetes/manager.go
```diff
@@ -150,6 +150,11 @@ func (m *Manager) HandleResources(
 		return nil, err
 	}
 
+	err = kuberesources.EnsureBinderRoleBinding(ctx, c, ns, sa.Name)
```
gman0: Previously, `kube-binder` RoleBinding was created here:

https://github.com/kube-bind/kube-bind/blob/d15b44c34d7d5844482cdb59ff6256912acc9c69/backend/controllers/servicenamespace/servicenamespace_reconcile.go#L251-L275

IMHO it makes sense to move it to here to the `/bind` handler, since this is the place where `kube-binder` ClusterRole as well as the `kube-bind-<Cluster>` namespace is created.

backend/controllers/serviceexportrbac/rbac_reconciler.go
```diff
@@ -0,0 +1,423 @@
...
+		{
+			Kind:      "ServiceAccount",
+			Namespace: export.Namespace,
+			Name:      kuberesources.ServiceAccountName,
+		},
+	}
+
+	errs = append(errs,
+		r.ensureAggregatingClusteRoleAndClusterRoleBinding(ctx, client, cache, kubebinderSubject),
+		// We always use cluster-scoped RBACs for exported resources.
```
gman0, line 177: RBACs for exported resources always use cluster-scoped permissions. This is how it was before too.

We could in theory use namespaced Roles & RoleBindings for resources too when `Isolation==Namespaced && Scope==Namespaced`.

backend/controllers/serviceexportrbac/rbac_controller.go
```diff
@@ -0,0 +1,222 @@
...
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *APIServiceExportRBACReconciler) SetupWithManager(mgr mcmanager.Manager) error {
+	return mcbuilder.ControllerManagedBy(mgr).
+		For(&kubebindv1alpha2.APIServiceExport{}).
+		Watches(
```
mjudeikis, line 180: Fo we need `Owns` here ?

backend/kubernetes/manager.go
```diff
@@ -150,6 +150,11 @@ func (m *Manager) HandleResources(
...
+	if err != nil {
+		return nil, err
+	}
```
ntnn:
```suggestion
	if err := kuberesources.EnsureBinderRoleBinding(ctx, c, ns, sa.Name); err != nil {
		return nil, err
	}
```

backend/controllers/serviceexportrbac/rbac_reconciler.go
mjudeikis: does this means that if I have prefix isolation, and claim secrets, I get access to all secrets in the cluster?

ntnn (commented):
Otherwise lgtm, but leaving final approve to @mjudeikis

backend/controllers/serviceexportrbac/rbac_reconciler_test.go
```diff
@@ -0,0 +1,867 @@
...
+			compareMap(t, "ClusterRole", tt.expectedState.clusterRoles, tt.st.clusterRoles)
+			compareMap(t, "ClusterRoleBinding", tt.expectedState.clusterRoleBindings, tt.st.clusterRoleBindings)
+			compareMap(t, "Role", tt.expectedState.roles, tt.st.roles)
+			compareMap(t, "RoleBinding", tt.expectedState.roleBindings, tt.st.roleBindings)
+		})
+	}
+}
+
+func compareMap[V any](t *testing.T, kind string, a, b map[string]V) {
+	require.Equal(t, slices.Sorted(maps.Keys(a)), slices.Sorted(maps.Keys(b)), "got unexpected entries after %s reconciliation", kind)
```
gman0, line 853: Should call `t.Helper()`.

backend/controllers/serviceexportrbac/rbac_reconciler.go
gman0: @mjudeikis that decision (whether konnector has cluster- or namespace-scoped access for claimed resources) is based on:

**Resource claim scope** | **Informer scope** | **What happens**
--- | --- | ---
cluster | cluster | Creates ClusterRole & ClusterRoleBinding
namespace | cluster | Creates ClusterRole & ClusterRoleBinding
--- | --- | ---
cluster | namespace | Creates ClusterRole & ClusterRoleBinding
namespace | namespace | **Creates Role & RoleBinding**

So konnector gets namespaced access to Secrets only if the informer scope itself is namespaced.

---

And this is the behavior we already have btw:

If `c.scope == kubebindv1alpha2.ClusterScope`:

https://github.com/kube-bind/kube-bind/blob/d84c3f0bf4b68bdb7038117f0d8a5992949f53a2/backend/controllers/servicenamespace/servicenamespace_reconcile.go#L100-L124

... otherwise (namespaced):

https://github.com/kube-bind/kube-bind/blob/d84c3f0bf4b68bdb7038117f0d8a5992949f53a2/backend/controllers/servicenamespace/servicenamespace_reconcile.go#L175-L189

gman0: My comment above ( https://github.com/kube-bind/kube-bind/pull/451#discussion_r2740899051 ) was just a suggestion that we could add (yet another :/) mode of operation -- this time for exported resources (`APIServiceExport.Spec.Resources`), where also backend isolation would be considered.

Right now exported resources always get cluster-scoped access, unconditionally.

backend/controllers/servicenamespace/servicenamespace_reconcile.go
```diff
@@ -81,327 +73,9 @@ func (c *reconciler) reconcile(ctx context.Context, client client.Client, cache
...
-					rb.RoleRef = rbacv1.RoleRef{Kind: "Role", Name: name, APIGroup: "rbac.authorization.k8s.io"}
-					if err := client.Update(ctx, rb); err != nil {
-						return fmt.Errorf("failed to update RoleBinding %s: %w", name, err)
-					}
-				}
-			}
-		}
-	}
-
-	if err := c.cleanupRBACResources(ctx, client, cache, sns, apiServiceExports); err != nil {
```
cnvergence, line 244: are we dropping cleanup of RBAC resources completely?

backend/controllers/serviceexportrbac/rbac_reconciler.go
```diff
@@ -0,0 +1,421 @@
...
+		},
+	}
+
+	clusterRole, err := r.getClusterRole(ctx, cache, name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if clusterRole == nil {
+		err = r.createClusterRole(ctx, client, &expectedClusterRole)
+		if err != nil && apierrors.IsAlreadyExists(err) {
```
cnvergence:
```suggestion
		if err != nil && !apierrors.IsAlreadyExists(err) {
```

backend/controllers/serviceexportrbac/rbac_reconciler.go
```diff
@@ -0,0 +1,421 @@
...
+
+	expectedRole := rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Rules: rules,
+	}
+
+	role, err := r.getRole(ctx, cache, namespace, name)
```
cnvergence:
```suggestion
role, err := r.getRole(ctx, cache, namespace, name)
if err != nil {
	if !apierrors.IsNotFound(err) {
		return err
	}
	if err := r.createRole(ctx, client, &expectedRole); err != nil && !apierrors.IsAlreadyExists(err) {
		return err
	}
	role = &expectedRole
}
```
let's maybe simplify it to something like

backend/controllers/serviceexportrbac/rbac_reconciler.go
mjudeikis: Can we add the table to the docs somewhere?

backend/controllers/serviceexportrbac/rbac_controller.go
gman0: Actually I've ended up removing `Controller: true` in the owner reference ( https://github.com/kube-bind/kube-bind/pull/451/changes/c4472ee403ca294b388a7c10e5a7429ed0f91a1a ).

I'm not sure `Owns()` would do much: the owner is the Namespace object, so we wouldn't get notified :/

backend/controllers/servicenamespace/servicenamespace_reconcile.go
gman0: All created RBAC-related objects have the owner set to `kube-binder-<Consumer's cluster name>`. So when that namespace is gone, the GC will do the clean up.

backend/controllers/serviceexportrbac/rbac_reconciler.go
gman0: I've moved this whole mess into `getOrCreate` function: https://github.com/kube-bind/kube-bind/pull/451/changes/5d08fa16563e6a797c502394e3ef3e0a41fd4cea

backend/controllers/serviceexportrbac/rbac_reconciler.go
gman0: @mjudeikis  Added in https://github.com/kube-bind/kube-bind/pull/451/changes/27f95330a3f950d3807583ad739aed415ee6dfd6

backend/controllers/servicenamespace/servicenamespace_reconcile.go
gman0: Do you think it's not enough? It's not perfect, I agree: there is one caveat in that if a consumer un-binds an export, they'd retain their RBACs until the ClusterBinding goes away (and with that the whole namespace, which cleans up the RBACs).

The question is if this actually does have any impact, and I'd hazard to say no, because the consumer/konnector could gain these permissions again by simply re-binding the service export.

backend/controllers/serviceexportrbac/rbac_controller.go
gman0: So that means we don't get notifications on RBACs at all. This is a bit meh, on the other hand we didn't have that before either.

We could add it if it's necessary -- maybe in a follow-up PR? If a care-free admin messes up the RBACs they could trigger a reconcile by nudging the APIServiceExport or APIServiceNamespace so it's not *that* critical...

PR #455, opened 2026-02-04
feat: add OwnerReferences to BoundSchema

### LINKED ISSUE #297: Add BoundSchema lifecycle policies
We need to think, what todo with BoundSchema lifecycle. Meaning that user should be able to specify if they want schema to be auto-updated, or always stay the same. 

This probably should be part of the provider configuration as they know how this could work with the service they are providing.

cnvergence:
## Summary

Setting OwnerReferences from BoundSchemas to their APIServiceExport, enabling automatic cleanup via Kubernetes garbage collection when the export is deleted.

## What Type of PR Is This?

/kind feature
## Related Issue(s)

Fixes #297 

## Release Notes

```release-note
Setting OwnerReferences from BoundSchemas to their APIServiceExport
```

### DIFF
--- a/backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
+++ b/backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
@@ -189,6 +189,10 @@ func (r *reconciler) createOrUpdateBoundSchema(ctx context.Context, cl client.Cl
 		return nil
 	}
 
+	// If namespaced isolation is configured for cluster-scoped objects,
+	// we need to rewrite the BoundSchema's scope accordingly. For all
+	// other isolation strategies, as well as for namespaced schemas,
+	// no changes are necessary.
 	if desired.Spec.Scope == apiextensionsv1.NamespaceScoped && r.isolation == kubebindv1alpha2.IsolationNamespaced {
 		desired.Spec.Scope = apiextensionsv1.ClusterScoped
 	}


backend/controllers/serviceexportrequest/serviceexportrequest_reconcile.go
```diff
@@ -75,15 +76,54 @@ func (r *reconciler) reconcile(ctx context.Context, cl client.Client, cache cach
 		return fmt.Errorf("failed to ensure APIServiceNamespaces: %w", err)
 	}
 
-	// TODO(mjudeikis): we could potentially add finallizer to APIServiceExport above or "adopt" boundschemas
-	// with owner references once export is created.
-	// https://github.com/kube-bind/kube-bind/issues/297
+	if err := r.setBoundSchemaOwnerRefs(ctx, cl, cache, req); err != nil {
+		conditions.SetSummary(req)
+		return fmt.Errorf("failed to adopt BoundSchemas: %w", err)
+	}
```
xrstf: For migrating existing setups I can see this being useful, but in general I think the ownerRef should just be set during ensureBoundSchemas(), instead of first creating the BoundSchemas and then immediately updating them afterwards.

mjudeikis: Agree, we even have this in the code too. This feels like wrong place. ensureResourceSchema should be more like "Create or Update if exist" (Update just ownerRef as we need wider lifecycle polices - do we update schemas or leave them drift)
```
				// TODO(mjudeikis): https://github.com/kube-bind/kube-bind/issues/297
				if obj != nil {
					continue
				}
```

cnvergence: Yes, I wanted to do it, but thought about a simpler way to do it.
We create APIServiceExports after the ensureBoundSchemas call, maybe we can include this logic in the ensureServiceExports or just refactor this.

PR #531, opened 2026-04-22
Always resolve the latest release version for chart and image default

cnvergence:
## Summary

Fix the issue where providing a custom tag leads to the deployment of the helm chart with a fallback version.
Also, bump the latest failback version to 0.7.1.
Restore old values to be compatible with older chart versions.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fixed the issue where providing a custom tag led to the deployment of the helm chart with a fallback version.
```


xrstf:
Isn't the BackendTag option now completely ignored?

cnvergence:
It is not, but I have checked that we need to keep old values for backward compatibility

mjudeikis (approved):

PR #546, opened 2026-05-18
Fix helm chart image version

mjudeikis:
## What Type of PR Is This?

Make the default image tag tolerate both v-prefixed and unprefixed `Chart.appVersion` by stripping any leading v and always prepending one. Fixes installs failing with `ImagePullBackOff` when `appVersion` is published without the v prefix (e.g. `0.8.1` (how helm charts does it) while images are tagged v0.8.1.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/helm-build.sh
+++ b/hack/helm-build.sh
@@ -33,7 +33,7 @@ for chart_dir in deploy/charts/*/; do
 
       cp "${chart_dir}Chart.yaml" "${chart_dir}Chart.yaml.bak"
       sed -i.tmp "s/^version:.*/version: $CHART_VERSION/" "${chart_dir}Chart.yaml"
-      sed -i.tmp "s/^appVersion:.*/appVersion: $CHART_VERSION/" "${chart_dir}Chart.yaml"
+      sed -i.tmp "s/^appVersion:.*/appVersion: v$CHART_VERSION/" "${chart_dir}Chart.yaml"
       rm -f "${chart_dir}Chart.yaml.tmp"
 
       "$HELM" package "$chart_dir" --version "$CHART_VERSION" --destination ./bin/


cnvergence (dismissed):

xrstf:
> when appVersion is published without the v prefix (e.g. 0.8.1 (how helm charts does it) while images are tagged v0.8.1.

Why? We control the appVersion, no? Why do we make the chart resilient against ourselves here? I am so confused... Why would the chart ever be published with a broken appVersion field?

cnvergence:
thought it was for the dev setup, now I see that latest tagged helm chart has a wrong appVersion, we should fix it

mjudeikis:
It might be now knowing helm again. But helm charts (OCI charts) dont accepts `v` prefix so helm is always `0.8.1` while our image tags are `v0.8.1`. And when helm is published I think its stomps appVersion withou `v` prefix. 

Maybe its issue with helm publishing? 

> now I see that latest tagged helm chart has a wrong appVersion

helm complains if OCI helm image is with `v` :/

cnvergence:
well, I ran into that when running development image and helm chart

mjudeikis:
So either this or we need to make sure helm charts stays without `v` and appversion is with `v` prefix (if possible)

xrstf:
Helm chart's `version` fields must be a valid semver, without the leading `v` prefix. But `appVersion` can be anything and depends on the wrapped application. For Minio you'd have `appVersion: "RELEASE-2026...."`, for example.

hack/helm-build.sh
```diff
@@ -33,7 +33,7 @@ for chart_dir in deploy/charts/*/; do
 
       cp "${chart_dir}Chart.yaml" "${chart_dir}Chart.yaml.bak"
       sed -i.tmp "s/^version:.*/version: $CHART_VERSION/" "${chart_dir}Chart.yaml"
-      sed -i.tmp "s/^appVersion:.*/appVersion: $CHART_VERSION/" "${chart_dir}Chart.yaml"
+      sed -i.tmp "s/^appVersion:.*/appVersion: v$CHART_VERSION/" "${chart_dir}Chart.yaml"
```
xrstf, line 36: I vote for a cleanup:

1. Rename `CHART_VERSION` in the `image` workflow to `APP_VERSION`. Also do not strip the leading `v` off it anymore (i.e. instead of running this script with `CHART_VERSION=0.8.1`, it should be run with `APP_VERSION=v.0.8.1`).
2. Strip the leading `v` here in this script to form a valid `version` for the Helm chart.
3. Replace `version` and `appVersion` in the Chart.yaml with `0.0.0` to indicate that their values are entirely irrelevant, should never be touched and would be overwritten with every kube-bind release anyway. Ideally a small comment `# updated by helm-build.sh during the build` is placed somewhere in there, too.

My reasoning is that my brain finds the thought

> the chart version is used as the app version

more confusing than

> the app version is also used as the chart version

^^ Especially since we're not actually maintining a dedicated version just for the chart, we derive it from the appVersion (the Git tag).

cnvergence: We could do this, wdyt @mjudeikis?

mjudeikis: Feel free to contribute. I dont see this complex, so if you have ideas how to make it better - take a stab.

cnvergence:
closing in favor of https://github.com/kube-bind/kube-bind/pull/554

PR #551, opened 2026-05-19
Fix packages retention in gh action image

cnvergence:
## Summary

It seems like we have been trying to clean-up non existent package 'kube-bind', old gh action did not report any issue.
This PR sets it for both backend and konnector packages.

## What Type of PR Is This?

/kind cleanup
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -83,7 +83,14 @@ jobs:
 
     - uses: actions/delete-package-versions@e5bc658cc4c965c472efe991f8beea3981499c55 # v5.0.0
       with:
-        package-name: '${{ github.event.repository.name }}'
+        package-name: 'konnector'
+        package-type: 'container'
+        min-versions-to-keep: 10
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yaml
```diff
@@ -83,7 +83,21 @@ jobs:
...
+      with:
+        package-name: 'backend'
+        package-type: 'container'
+        min-versions-to-keep: 10
+        delete-only-pre-release-versions: "true"
+
+    - uses: actions/delete-package-versions@e5bc658cc4c965c472efe991f8beea3981499c55 # v5.0.0
+      with:
+        package-name: 'charts/backend'
         package-type: 'container'
```
xrstf, line 101: Is this truly a container?

cnvergence: Helm charts pushed to GHCR via OCI are actually stored as container types

xrstf (approved):

PR #560, opened 2026-06-10
Add v2 proposals

mjudeikis:
## Summary

We have been integrating kube-bind with external systems, and we now understand what works and what does not for users to use it in a smooth, frictionless way. 

I tried to draft an initial v2 proposal for Core and the remaining features. 

The idea is to simplify the flow, make the core directory usable for the rest of the features, so one can use it separately as "syncer only".  Historically, some of these features got globbed together, and in some cases, overengineered. 

In addition, during the initial bootstrap, https://github.com/multicluster-runtime/multicluster-runtime did not exist. Now it does. 

This serves as the initial discussion started. Please review and let me know wdyt? 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
V2 proposal document 
```

### DIFF
--- a/docs/proposals/v2-extended.md
+++ b/docs/proposals/v2-extended.md
@@ -1,0 +1,262 @@
+# kube-bind v2 Extended: Backend API, CLI, UI
+
+* Status: **DRAFT — for iteration**
+* Authors: @mjudeikis
+* Date: 2026-06-10
+* Builds on: [v2-slim-core.md](v2-slim-core.md) (Proposed))
+
+## Summary
+
+The v2 core contract is up for discussion: a binding is one `kubectl apply` of a Secret +
+`Connection` + bindings, consumed by the konnector, with zero kube-bind components on
+the provider. This proposal designs everything *around* that contract — the optional
+service layer that answers the questions the core deliberately doesn't:
+
+* **Who are you?** (auth: OIDC, sessions)
+* **What may you have?** (catalog: curated offerings on top of raw exported APIs)
+* **Here are your credentials.** (issuer: per-consumer SA/RBAC/kubeconfig, tenancy)
+* **Here is your bundle.** (gateway: HTTP API whose terminal output is the one-apply file)
+* **You stopped coming.** (reaper: GC keyed off the core's Lease)
+* **Make it pleasant.** (CLI, UI)
+
+The defining rule, inherited from the core: **every path through this layer terminates
+in the same one-apply bundle.** The service layer negotiates; it never syncs. If the
+backend is deleted the day after binding, sync is unaffected.
+
+```
+            provider cluster                                consumer cluster
+ ┌────────────────────────────────────┐
+ │  extended layer (this proposal)    │
+ │  ┌──────────┐ ┌────────┐ ┌──────┐  │     bundle
+ │  │ gateway  │ │ issuer │ │reaper│  │  (one-apply file)      ┌───────────┐
+ │  │ auth, UI │ │ creds, │ │ Lease│  │ ──────────────────────▶│ konnector │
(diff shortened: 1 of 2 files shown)


docs/proposals/v2-slim-core.md
```diff
@@ -0,0 +1,643 @@
...
+    categories, `namespaced`, verbs, and the served versions of the group;
+  * **`/openapi/v3/apis/<group>/<version>`** → the structural schema per version,
+    including `x-kubernetes-*` extensions and defaults;
+  * subresource detection (status/scale) from the OpenAPI paths.
+
+  Known fidelity limits, stated honestly: CEL validation rules and other
+  server-side-only constraints may not round-trip through OpenAPI — the consumer-side
+  CRD can be *looser* than the provider's real validation, and the provider remains the
+  enforcing side (a consumer object that passes locally can still be rejected upstream;
+  that surfaces as a per-object sync condition, not silent loss).
```
ntnn, line 326: Thinking out loud - couldn't this be fixed with a validation webhook? Install kbind as valdiation webhook on the consumer and proxy that to the provider or make a dry-run apply or something? That way the provider can still do the full validation. would be hacky though and the tradeoff mentioned would be ok i think.

mjudeikis: I don't want to have validation webhooks. In general, webhooks in any cluster require way more care and support, and from experience, you need to "have full access to the cluster" to stay on top of it. + it required a certificate pair, which adds dependency on cert-manager or another cert issuer. It just very hard to support in wide range of cluster configurations with wide range of setup permutations. 

In general, it's easier to ship lightweight components if they do not have webhooks.  

All this with the exception of CEL-based webhooks https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions but it has limitations what it can do.

docs/proposals/v2-extended.md
```diff
@@ -0,0 +1,280 @@
...
+  delete; deletion requires explicit enablement).
+
+### 6. CLI (`kubectl bind`)
+
+Thin client over the gateway; everything it does is reproducible by hand:
+
+```sh
+kubectl bind login https://mangodb.example.com        # auth, cache token
+kubectl bind catalog                                  # list Exports/Collections
+kubectl bind mangodb                                  # bind an Export:
```
sttts: why not `kubectl bind export mangodb`? Mixing commands and custom export names seems wrong.

docs/proposals/v2-slim-core.md
```diff
@@ -0,0 +1,643 @@
...
+   `None`) × two informer scopes (`Cluster`, `Namespaced`) × `APIServiceNamespace`
+   namespace-mapping objects. Cluster-scoped resources can be *converted* to namespaced on
+   the provider, names get rewritten, namespaces get re-mapped. This is the single largest
+   source of complexity in the syncer (the whole
+   [isolation/](../../pkg/konnector/controllers/cluster/serviceexport/isolation/) strategy
+   layer plus `APIServiceNamespace` round-trips in every reconcile), and the source of the
+   most confusing API fields (`informerScope` vs `isolation` vs deprecated
+   `clusterScopedIsolation` vs CRD `scope`).
+
+2. **The object zoo.** A bind today touches up to ten CRDs: `APIServiceExportTemplate`,
```
cnvergence, line 72: +1

cnvergence: Would be nice to see what apis go into v2 and which doesn't, like a table

cnvergence (approved):
LGTM, we can start moving towards v2

